PR #16, opened 2019-03-06
feat(scripts): add scripts module

MaxPostema:
#### Checklist
- [x] Code reviewed
- [ ] Updated flow typing


packages/scripts/src/store/getters.ts
```diff
@@ -0,0 +1,7 @@
+import State from '@/types/state'
+
+export default {
```
fdlk: I don't think you actually need these.

packages/scripts/README.md
```diff
@@ -0,0 +1,39 @@
+# molgenis-scripts
```
connoratrug: please add short description of functionality, maybe point to docs for further info

packages/scripts/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.15",
+    "@fortawesome/free-regular-svg-icons": "^5.7.2",
+    "@fortawesome/free-solid-svg-icons": "^5.7.2",
+    "@fortawesome/vue-fontawesome": "^0.1.5",
+    "@molgenis/molgenis-api-client": "^3.1.0",
+    "@molgenis/molgenis-ui-form": "^2.1.0",
+    "vue": "^2.6.8",
+    "vue-class-component": "^7.0.1",
```
connoratrug: i do not think this is used, please remove if unused

packages/scripts/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "vue-property-decorator": "^7.3.0",
```
connoratrug: same as for class component

packages/scripts/public/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <title>molgenis-scripts</title>
+    <link rel="stylesheet" href="/css/bootstrap-4/bootstrap-molgenis-blue.min.css?tcBVgw" type="text/css" id="bootstrap-theme">
```
connoratrug: would suggest tot replace with bootstrap

packages/scripts/src/App.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <div id="app">
+    <b-alert v-if="showError" variant="danger" show dismissible>
+      Error: Can't connect to molgenis.
```
connoratrug: use i18n

packages/scripts/src/App.vue
```diff
@@ -0,0 +1,30 @@
...
+      <router-link to="/login" tag="button">Sign in</router-link>
```
connoratrug: use i18n

packages/scripts/src/App.vue
```diff
@@ -0,0 +1,30 @@
...
+  },
+  created () {
+    this.$store.dispatch('requestScripts').catch(() => {
+      this.showError = true
+    })
+  }
+}
+</script>
+
+<style>
```
connoratrug: remove empty tag

packages/scripts/src/components/CodeEditor.vue
```diff
@@ -0,0 +1,86 @@
...
+  <div>
+    <div class="position-relative">
+      <form-component v-if="showField"
+                      id="CodeEditor"
+                      :formFields="codeEditorFields"
+                      :initialFormData="codeEditorData"
+                      :formState="codeEditorState"
+                      :options="options"
+                      @valueChange="onValueChanged"></form-component>
+      <span id="in-corner"><slot></slot></span>
```
connoratrug: i do not understand this id

packages/scripts/src/components/CodeEditor.vue
```diff
@@ -0,0 +1,86 @@
...
+      },
+      codeEditorState: {},
+      parameters: []
+    }
+  },
+  methods: {
+    onValueChanged (event) {
+      this.codeEditorData = event
+      const code = event.content
+      const parameter = /\${(\w+)}/g // Select ${name} from code
```
connoratrug: suggest to rename to parameterRegEx and move to separate , tested function

packages/scripts/src/components/CodeEditor.vue
```diff
@@ -0,0 +1,86 @@
...
+      let result
+      let list = []
+      while ((result = parameter.exec(code)) !== null) {
```
connoratrug: i dont get the while loop , suggest to move the extraction to separate function.

packages/scripts/src/components/ExecuteButton.vue
```diff
@@ -0,0 +1,64 @@
...
+        })
+      } else {
+        this.run()
+      }
+    },
+    run () {
+      if (this.parameters.length > 0) {
+        this.showModal = true
+      } else {
+        window.open('/scripts/' + this.name + '/start', '_blank')
```
connoratrug: as a user i really do not like this part , but i guess it is the same as it was before , so we should keep it "

packages/scripts/src/components/RunModal.vue
```diff
@@ -0,0 +1,47 @@
...
+    },
+    value (changedValue) {
+      this.showModal = changedValue
+    }
+  },
+  methods: {
+    run () {
+      this.showModal = false
+      let url = '/scripts/' + this.name + '/start?'
+      this.parameters.forEach((key, index) => {
```
connoratrug: use map, reduce instead of foreach

packages/scripts/src/main.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { faEdit, faTrash, faPlay } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+
+Vue.config.productionTip = false
+
+library.add(faEdit, faTrash, faPlay)
+Vue.component('font-awesome-icon', FontAwesomeIcon)
+Vue.use(BootstrapVue)
```
connoratrug: have you considered not using bootstrap vue ? :)

packages/scripts/src/types/state.ts
```diff
@@ -0,0 +1,10 @@
+// export const enum ReadyState { loading = 1, failed, ready }
+export enum ReadyState { loading = 'loading', failed = 'failed', ready = 'ready' }
+
+export interface State{
+  scripts: any[]
+  scriptTypes: string[]
```
connoratrug: could fix this to the types we have

packages/scripts/src/types/state.ts
```diff
@@ -0,0 +1,10 @@
+// export const enum ReadyState { loading = 1, failed, ready }
+export enum ReadyState { loading = 'loading', failed = 'failed', ready = 'ready' }
+
+export interface State{
+  scripts: any[]
```
connoratrug: is this really any ?

packages/scripts/src/views/ListScripts.vue
```diff
@@ -0,0 +1,141 @@
...
+    },
+    createNewScript () {
+      this.$router.push({ name: 'newscript' })
+    }
+  }
+})
+</script>
+
+<style scoped>
+  >>> .compact {
```
connoratrug: why are these details needed , is this going to work work with the theming ?

packages/scripts/tests/e2e/specs/test.js
```diff
@@ -0,0 +1,14 @@
+// For authoring Nightwatch tests, see
+// http://nightwatchjs.org/guide#usage
+
+module.exports = {
+  'default e2e tests': browser => {
+    browser
```
connoratrug: please update this to basic test , would be even nice to test basic functions 
- list
- add 
- edit

packages/scripts/vue.config.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  css: {
+    extract: {
+      filename: 'css/scripts/[name].css',
+      chunkFilename: 'css/scripts/[name].css'
+    }
+  },
+  'devServer': {
+    'proxy': {
+      '^/api': {
```
connoratrug: i think you can combine these

packages/scripts/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### Compiles and minifies for production
+```
+yarn run build
+```
+
+### Run your tests
+```
+yarn run test
```
connoratrug: this command does not work

packages/scripts/src/components/CodeEditor.vue
MaxPostema: Changed to class, used to target and place button via css

packages/scripts/src/components/RunModal.vue
MaxPostema: I never used reduce before, but it looks much better now.

marikaris:
the example in the docs doesn't work anymore: https://molgenis.gitbooks.io/molgenis/content/user_documentation/scripts/guide-scripts.html

I pasted this:
```R
library('RCurl')
library('methods')
library('bitops')
eval(expr = parse(text = getURL("https://molgenis56.target.rug.nl/molgenis.R")))

samples <- molgenis.get("SampleAse", q="SNP_ID==${snp_id}")
jpeg('${outputFile}')

max <- max(samples$Ref_Counts, samples$Alt_Counts)

plot(samples$Ref_Counts, samples$Alt_Counts, xlim = c(0, max), ylim = c(0, max), xlab='Reference Allele', ylab='Alternative Allele', main = 'Allele-Specific Expression for ${snp_id}')
lines(c(0,max), c(0, max))
dev.off()
```

MaxPostema:
@marikaris 
> the example in the docs doesn't work anymore
> ...

This is fixed with "fix(scripts): fix no longer storing molgenisToken and outputFile"

packages/scripts/src/main.ts
MaxPostema: Yes i have, still going to use it though. 😋

packages/scripts/src/components/CodeEditor.vue
MaxPostema: Moved to 'findParameters' function

packages/scripts/vue.config.js
MaxPostema: I tried a few things. But the redirect to the job handler keeps breaking every time i change this.

packages/scripts/tests/e2e/specs/test.js
MaxPostema: I removed the e2e test for now. I think this is something to think about and have a good talk about. I would like the test, but i do not like to have molgenis in the background to do so. Mocking a api in a e2e is not best practice.

packages/scripts/src/types/state.ts
MaxPostema: changed to object.

packages/scripts/src/types/state.ts
MaxPostema: Rebuilding the complete API structure into a complex datatype feels like a lot of over engineering just to remove the "any". I think that a more 'save' type will also take a lot of time then something about the api changes. (like adding a new field)

packages/scripts/src/views/ListScripts.vue
MaxPostema: `.compact{}` is used to make sure that the row is as small as it can be.
`tr {}` I do enable bootstraps 'hover' table style, but i also want the cursor to change. Not just the hovered row to lit up.
`.badge` the smaller badge i use for the variables don't sit nice, this will make it look better.

All changes are scoped and i don't see any theming problems.

packages/scripts/tests/e2e/specs/test.js
MaxPostema: working e2e test added

packages/scripts/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@molgenis/scripts",
+  "version": "0.1.0",
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint",
+    "build-watch": "vue-cli-service build --watch",
+    "e2e": "vue-cli-service test:e2e",
```
connoratrug: why are these not the default vue cli targets ? suggest to use is the defaults unless this is not possible

MaxPostema: The default Lerna/Jenkinsfile in the root expects this calling convention.

packages/scripts/src/types/state.ts
connoratrug: No changes on the api response do not necessarily directly impact the model, if they do then it would be good to know about it. I do not suggest you 'Rebuilding the complete API structure', i suggest to add type information to the  frontend model, to communicate the model structure to other devs ( or your future self :) )

PR #39, opened 2019-04-10
fix(one-click-importer): Set height of file input to 100%

marikaris:
Set height of file input in one click importer to 100% to override the faulty bootstrap theme
calculation. We chose to keep the current file input rather than replace it by the bootstrap one
because we like the browser translations, those are lost when the bootstrap variant is used.

Fixes molgenis/molgenis#8368

#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/one-click-importer/src/components/OneClickImporter.vue
+++ b/packages/one-click-importer/src/components/OneClickImporter.vue
@@ -7,7 +7,7 @@
             <label for="file-input">{{ 'import-input-label' | i18n }}</label>
             <input
               id="file-input"
-              class="form-control"
+              class="form-control molgenis-file-input"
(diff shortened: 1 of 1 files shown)


packages/one-click-importer/src/components/OneClickImporter.vue
```diff
@@ -97,6 +97,10 @@
     padding-top: 0.5em;
     font-size: smaller;
   }
+
+  #file-input {
```
connoratrug: Could you do a class ?, for example '.mg-file-input'  , that way we can keep classes for styling and use id's for linking to js and form stuff.  and maybe add a comment line explaining this is a work around ?

PR #43, opened 2019-04-11
fix(scripts): fix #8384 use different extractor based on script type

connoratrug:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/scripts/tests/unit/service/parameterService.spec.ts
+++ b/packages/scripts/tests/unit/service/parameterService.spec.ts
@@ -1,17 +1,29 @@
 import { findParameters } from '@/service/parameterService'
 
 describe('service/parameterService', () => {
-    describe('findParameters', () => {
-        it('Find non magma parameter', () => {
+    describe('FindParameters for non Magma types (freemarker style)', () => {
+        it('should find \'koe\' in \'bla bla bla ${koe} bla bla\'', () => {
             expect(findParameters('bla bla bla ${koe} bla bla', 'test')).toEqual(['koe'])
         })
+    })
 
-        it('Find magma parameter', () => {
+    describe('FindParameters for Magma script', () => {
+        it('should find \'myValue\' in \'bla bla bla $("myValue") bla bla\'', () => {
             expect(findParameters('bla bla bla $("myValue") bla bla', 'JavaScript (Magma)')).toEqual(['myValue'])
         })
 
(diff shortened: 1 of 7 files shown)


packages/scripts/tests/unit/service/parameterService.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+describe('service/parameterService', () => {
+    describe('findParameters', () => {
+        it('Find non magma parameter', () => {
+            expect(findParameters('bla bla bla ${koe} bla bla', 'test')).toEqual(['koe'])
+        })
+    })
+
+    describe('findParameters', () => {
+        it('Find magma parameter', () => {
+            expect(findParameters("bla bla bla $('myValue') bla bla", 'JavaScript (Magma)')).toEqual(['myValue'])
```
fdlk: add (and support) a couple more interesting cases like `$("blah")` and `$('blah-di-blah')`

fdlk: See https://github.com/molgenis/molgenis/blob/master/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/NameValidator.java for an improvement over `\w`

fdlk:
After a lengthy discussion with @dennishendriksen :
There will definitely be cases where this regex approach will fail for magmascript and there may well be cases where the regex approach will fail for other script types

And we do not currently give the user a way to override our guesses. This is problematic.

packages/scripts/tests/unit/service/parameterService.spec.ts
```diff
@@ -5,11 +5,13 @@ describe('service/parameterService', () => {
...
-    })
 
-    describe('findParameters', () => {
         it('Find magma parameter', () => {
-            expect(findParameters("bla bla bla $('myValue') bla bla", 'JavaScript (Magma)')).toEqual(['myValue'])
+            expect(findParameters('bla bla bla $("myValue") bla bla', 'JavaScript (Magma)')).toEqual(['myValue'])
+        })
+
+        it('should not find params with single quotes in when type is JavaScript (Magma)', () => {
+            expect(findParameters("$('blah-di-blah')", 'JavaScript (Magma)')).toEqual([])
```
fdlk: blah-di-blah is a valid attribute name

PR #50, opened 2019-04-17
Merge pull request #43 from connoratrug/fix/8384-js-params

connoratrug:
Reverts molgenis/molgenis-frontend#45


packages/scripts/src/service/parameterService.ts
```diff
@@ -1,5 +1,5 @@
 const freeMarkerRegEx = /\${([a-zA-Z0-9-#]+)}/g // Match foo from ${foo}
-const magmaRegEx = /\$\(\"([a-zA-Z0-9-#]+)\"\)|$\(\'([a-zA-Z0-9-#]+)\'\)/g // Match foo from $('foo') or $("foo")
+const magmaRegEx = /\$\(\"([a-zA-Z0-9-#]+)\"\)|\$\(\'([a-zA-Z0-9-#]+)\'\)/g // Match foo from $('foo') or $("foo")
```
connoratrug: please add a test case

fdlk: I've fixed the broken test that was supposed to test single quotes

connoratrug:
closed in favour of https://github.com/molgenis/molgenis-frontend/pull/54

comments 

```
Running into a bug where the parameter list does not get updated when I change the script type from and to magmascript, both in the yarn serve and in the preview
Linter complains about quotes and thinks that the regex is excaped more than needed.
```

are addressed in this PR

PR #72, opened 2019-06-04
Task/publish to npm

fdlk:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -121,19 +121,6 @@ pipeline {
                 }
             }
         }
-        stage('Release canary: [ master ]'){
-            when {
-                branch 'master'
-            }
-            steps {
-                lock("Tags"){
-                    sh "git fetch --tags"
-                    container('node') {
-                        sh "yarn lerna publish --canary"
-                    }
-                }
-            }
-        }
         stage('Release: [ master ]') {
             when {
                 allOf {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,7 @@ pipeline {
                 }
             }
         }
-        stage('Push to registries [ PR ]') {
+        stage('Build container serving the artifacts [ PR ]') {
             when {
                 changeRequest()
             }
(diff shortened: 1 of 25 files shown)


Jenkinsfile
```diff
@@ -58,8 +58,8 @@ pipeline {
                 container (name: 'kaniko', shell: '/busybox/sh') {
                     sh "#!/busybox/sh\nmkdir -p /root/.docker/"
```
connoratrug: maybe add a few line describing why this is happening

connoratrug:
@fdlk 

maybe this is a bit more elegant:
```
rsync -av packages/ docker/dist/ --exclude */node_modules
```

packages/scripts/vue.config.js
```diff
@@ -18,8 +18,6 @@ module.exports = {
     config.externals = {
       'bootstrap': 'bootstrap'
     }
-    config.output.filename = 'js/scripts/[name].js',
```
connoratrug: lines can be removed , were needed to have /scripts/ subpath

packages/scripts/vue.config.js
```diff
@@ -28,8 +26,8 @@ module.exports = {
   },
   css: {
     extract: {
-      filename: 'css/scripts/[name].css',
-      chunkFilename: 'css/scripts/[name].css'
+      filename: 'css/[name].css',
```
connoratrug: lines can be removed , sub path no longer needed

packages/scripts/vue.config.js
fdlk: I think otherwise we'll get a chunk hash in the name

fdlk:
> @fdlk
> 
> maybe this is a bit more elegant:
> 
> ```
> rsync -av packages/ docker/dist/ --exclude */node_modules
> ```

I don't think busybox has rsync

packages/scripts/vue.config.js
connoratrug: there is a separate , defidated flag for that

PR #96, opened 2019-08-08
feat(data-explorer): Add shoppingcart

MaxPostema:
Testing with the following link wil give some data to show:
http://localhost:8080/entity/it_emx_datatypes_TypeTest

#### Checklist
- [x] Code reviewed


packages/data-explorer/src/components/DataView/EntityCard.vue
```diff
@@ -0,0 +1,36 @@
+<template>
```
connoratrug: Do we keep the use of the term entity ? , maybe ExplorerCard or something more ui centric ?

packages/data-explorer/src/components/DataView/EntityCard.vue
```diff
@@ -0,0 +1,36 @@
...
+  <div class="card">
+    <shopping-button v-if="isShop" :isSelected="isSelected" @click.native="clickShop(id)"></shopping-button>
```
connoratrug: what is the function of the 'native' modifier , i cant find it in the vue docs

packages/data-explorer/src/components/DataView/EntityCard.vue
```diff
@@ -0,0 +1,36 @@
...
+    id: {
+      type: String,
+      required: true
+    },
+    isSelected: {
+      type: Boolean,
+      required: false,
+      default: () => false
+    },
+    clickShop: {
```
connoratrug: does the shop need to be explicit part of the Card ? would it not be better to use composition or slots ?

packages/data-explorer/src/components/DataView/EntityTableHeader.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
connoratrug: name , TableHeader ?

packages/data-explorer/src/components/DataView/EntityTableHeader.vue
```diff
@@ -0,0 +1,28 @@
...
+
+<script>
+export default {
+  name: 'EntityTableHeader',
+  props: {
+    header: {
+      type: Array,
+      required: true
+    },
+    isShop: {
```
connoratrug: same thing does the shop need to be part of the component ?

packages/data-explorer/src/components/DataView/EntityTableRow.vue
```diff
@@ -0,0 +1,38 @@
+<template>
```
connoratrug: name , TableRow ?

packages/data-explorer/src/components/DataView/EntityTableRow.vue
```diff
@@ -0,0 +1,38 @@
...
+  <tr>
+    <td v-if="isShop"><shopping-button :isSelected="isSelected" @click.native="clickShop(id)"></shopping-button></td>
```
connoratrug: same , shop seems to be leaking through whole of explorer

packages/data-explorer/src/components/DataView/EntityTableRow.vue
```diff
@@ -0,0 +1,38 @@
...
+<script>
+import ShoppingButton from '../Utils/ShoppingButton'
+export default {
+  name: 'EntityTableRow',
+  props: {
+    id: {
+      type: String,
+      required: true
+    },
+    data: {
```
connoratrug: data is very general , suggest to use 2 word combination for more descriptive  prop

packages/data-explorer/src/components/DataView/EntityTableRow.vue
```diff
@@ -0,0 +1,38 @@
...
+      type: Object,
+      required: true
+    },
+    isSelected: {
+      type: Boolean,
+      required: false,
+      default: () => false
+    },
+    clickShop: {
```
connoratrug: naming; 'handleShopButtonClick' , suggest using consistent naming schema

packages/data-explorer/src/main.ts
```diff
@@ -9,7 +9,9 @@ import i18n from '@molgenis/molgenis-i18n-js'
 Vue.config.productionTip = false
 
 Vue.use(i18n, {
-  namespace: ['filters'],
+  lng: 'en',
```
connoratrug: oke for now be need to think about getting this from backend ( maybe via context ?)

packages/data-explorer/src/types/ApplicationState.ts
```diff
@@ -3,11 +3,65 @@ export type Toast = {
   message: string
 }
 
+type DataApiResponseItem = {
```
connoratrug: suggest to split up server communication types ( maybe even share across apps) from true frontend app types

packages/data-explorer/src/views/EntityView.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <div class="mt-1 entity-table" v-if="activeEntityData && activeEntityData.items.length > 0 && entityMeta">
+    <div v-if="isShop && entitiesToShow.length === 0" class="alert alert-warning">
+      {{ 'dataexplorer_empty_shopping_cart' | i18n}}
+    </div>
+
+    <div class="card-columns" v-else-if="dataDisplayLayout==='cards'">
+      <entity-card
```
connoratrug: maby move to separate view components and use the dynamic component stuff instead of if else, then again this seems fine for now

packages/data-explorer/src/views/EntityView.vue
```diff
@@ -0,0 +1,125 @@
...
+        <div class="card-body">
+          <h5 class="card-title">{{entity.data.xstring}}</h5>
+          <p class="card-text">{{entity.data.xtext}}</p>
+        </div>
+
+      </entity-card>
+    </div>
+
+    <table v-else class="table">
+      <entity-table-header :header="Object.keys(entitiesToShow[0].data)" :isShop="isShop"></entity-table-header>
```
connoratrug: Suggest to replace complex header from function with named computed prop

packages/data-explorer/public/index.html
```diff
@@ -6,7 +6,24 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>data-explorer</title>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <!-- Molgenis stylesheet -->
+    <link rel="stylesheet" href="http://localhost:8080/css/bootstrap-4/bootstrap-molgenis-blue.min.css">
+    <style>
```
connoratrug: agree needs to move to general styling

packages/data-explorer/src/views/MainView.vue
```diff
@@ -25,25 +34,38 @@ import FiltersView from './FiltersView'
...
+      this.$store.commit('setActiveEntity', this.$route.params.entity)
+    }
   },
-  components: { FiltersView, DataView, ToastComponent }
+  components: { FiltersView, DataView, ToastComponent, FontAwesomeIcon }
 })
 </script>
 
 <style scoped>
+  .show-filters-button{
```
connoratrug: add space before '{'

packages/data-explorer/src/main.ts
MaxPostema: there is a sticky for that!

packages/data-explorer/src/views/EntityView.vue
MaxPostema: We merged this on purpose because the entitycards and entityrows are different views that use exactly the same methods and functions. We merged this to prevent code duplication.

packages/data-explorer/src/views/EntityView.vue
MaxPostema: Done. Note that this is only a temporary 'walking skeleton' table

packages/data-explorer/src/types/ApplicationState.ts
MaxPostema: cart will get its own view in an other pr

packages/data-explorer/public/index.html
```diff
@@ -6,7 +6,24 @@
...
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>data-explorer</title>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <!-- Molgenis stylesheet -->
+    <link rel="stylesheet" href="/css/bootstrap-4/bootstrap-molgenis-blue.min.css">
+    <style>
+
+      /* think of a way to changes filters styling? */
+      .flex-filter .card {
+        margin-bottom: 1rem !important;
```
marikaris: could you remove important?

PR #97, opened 2019-08-13
feat/renamed-menu-to-legacy-lib

sidohaakma:
#### Checklist
- [ ] Code reviewed


packages/context/src/App.vue
```diff
@@ -17,8 +17,8 @@ export default Vue.extend({
   },
   data () {
     return {
-      molgenisMenu: window.molgenisMenu,
-      showCookieWall: window.cookieWall
+      molgenisMenu: molgenisMenu,
```
connoratrug: why remove window ?, it makes it less magical , although the window is pretty magical ...

PR #114, opened 2019-08-29
feat(legacy-lib): add context library

fdlk:
#### Checklist
- [x] Code reviewed


packages/legacy-lib/vue.config.js
```diff
@@ -0,0 +1,18 @@
+const CopyPlugin = require('copy-webpack-plugin')
+
+// vue.config.js
+module.exports = {
+  filenameHashing: true,
```
connoratrug: i think this is the defualt

PR #133, opened 2019-09-13
chore: fix backend max body size

fdlk:
Allow big requests to the backend

#### Checklist
- [ ] Code reviewed


fdlk:
The PR still shows a 413 BUT that's because the request gets kicked by the proxy pod of the master.dev.molgenis.org deployment, not by the proxy pod

PR #141, opened 2019-09-30
meta to filter config

MaxPostema:
Note: 
- Only categoricals are added in this PR

#### Checklist
- [x] Code reviewed
- User documentation updated
- No warnings during install


packages/data-explorer/src/repository/metaFilterMapper.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }))
+
+  return {
+    definition: categoricals,
+    shown: shownFilters
+  }
+}
+
+const getOptions = async (href: String) => {
```
sidohaakma: I f**cked up. This needs to be a primitive "string" instead of "String"

PR #154, opened 2019-10-07
test(data-explorer): Update e2e test URLs

fdlk:
#### Checklist
- [x] Code reviewed
- User documentation updated
- No warnings during install
- Updated javascript typing

### DIFF
--- a/packages/data-explorer/tests/e2e/specs/test.js
+++ b/packages/data-explorer/tests/e2e/specs/test.js
@@ -34,7 +34,7 @@ module.exports = {
   },
   'should display custom card': browser => {
     browser
-      .url(process.env.VUE_DEV_SERVER_URL + 'TableWithCustomCard')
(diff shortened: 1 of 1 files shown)


packages/data-explorer/tests/e2e/specs/test.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
   },
   'should display custom card': browser => {
     browser
-      .url(process.env.VUE_DEV_SERVER_URL + 'TableWithCustomCard')
+      .url(process.env.VUE_DEV_SERVER_URL + '#/TableWithCustomCard')
```
connoratrug: i would think you would need to update the mock backend in the vue.config for this to work

connoratrug: > i would think you would need to update the mock backend in the vue.config for this to work

guess not: only used for data requests

PR #277, opened 2020-03-12
feat(data-explorer): Add search

connoratrug:
- Add search component
- Add search field to state
- Make it work together with the filters

Implementation uses the existing filter framework but adds 'search' filter to the list of filter via de data-explorer. In this way existing active-filter components can be used and existing rsql pipeline can be used. Search filter is a 'string' filter and use special rsql search ( with wild card) operator.

#### Checklist
- [ ] Code reviewed


connoratrug:
close in favour of https://github.com/molgenis/molgenis-frontend/pull/278

PR #278, opened 2020-03-16
Feat/#7190 add search

connoratrug:
- Add search component
- Add search field to state
- Make it work together with the filters

Implementation uses the existing filter framework but adds 'search' filter to the list of filter via de data-explorer. In this way existing active-filter components can be used and existing rsql pipeline can be used. Search filter is a 'string' filter and use special rsql search ( with wild card) operator.

#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/data-explorer/src/components/SearchComponent.vue
+++ b/packages/data-explorer/src/components/SearchComponent.vue
@@ -1,0 +1,61 @@
+<template>
+  <div class="input-group mb-3">
+    <input
+      type="text"
+      v-model="searchText"
+      class="form-control"
+      placeholder="Search"
+      aria-label="Search data"
+      aria-describedby="mg-data-explorer-search"
+    />
+    <div class="input-group-append">
+      <button
+        class="btn btn-outline-secondary"
+        type="button"
+        id="mg-data-explorer-search"
+        @click="handleSearchAction"
+      >
+        <font-awesome-icon icon="search"></font-awesome-icon>
+      </button>
+    </div>
+  </div>
+</template>
+
+<script>
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { faSearch } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+
+library.add(faSearch)
+
+export default {
+  name: 'SearchComponent',
+  components: { FontAwesomeIcon },
+  props: {
+    value: {
+      type: String,
+      required: false
+    }
+  },
+  data () {
+    return {
+      searchText: this.value
+    }
+  },
+  methods: {
+    handleSearchAction () {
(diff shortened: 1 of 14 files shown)


packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div class="input-group mb-3">
+    <input
+      type="text"
+      v-model="searchText"
+      class="form-control"
+      placeholder="Search"
```
MaxPostema: Maybe usei18n here?

packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -0,0 +1,61 @@
...
+      aria-label="Search data"
+      aria-describedby="mg-data-explorer-search"
+    />
+    <div class="input-group-append">
+      <button
```
MaxPostema: The rest of the interface (filters) react instantly. Having to push a button to search breaks my expectation. Maybe make search reactive?

packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -0,0 +1,61 @@
...
+        class="btn btn-outline-secondary"
+        type="button"
+        id="mg-data-explorer-search"
+        @click="handleSearchAction"
+      >
+        <font-awesome-icon icon="search"></font-awesome-icon>
+      </button>
+    </div>
+  </div>
+</template>
```
MaxPostema: feedback

I think that if the header items are a bit better aligned to the block below that the whole thing would look a bit less cluttered and more professional.

Even if we world only solve that stange ~5px margin, i think things would look cleaner.

packages/data-explorer/src/components/SearchComponent.vue
connoratrug: yes , but is not used elsewhere in data-explorer , suggest we skipt it for now and add it all at once. still some questions like do we add a new i18 namespace for explorer 2 or do we use the existing namespace , and do we add i18 data-explorer defaults to the war ( needs to done in molgenis repo i think)

packages/data-explorer/src/components/SearchComponent.vue
connoratrug: This is done by design, its search as opposed to typeahead ( filter style) , it's the same UX as the current data explorer, suggest to keep it for now and add it later as a feature if we think we want it ?

packages/data-explorer/src/components/SearchComponent.vue
connoratrug: Could you make a mockup in figma ( just edit the current one in molgenis ) to show what you suggest ? , i agree we need to clean this , but not yet sure on best layout , suggest to figure out design in figma and not in PR's or else it will take forever.

PR #280, opened 2020-03-25
Add create, delete and edit card/row controls

connoratrug:
- [x] Add add row/card button to toolbar
   - Add button with link to data-row-edit plugin
   - Do not show button in cart view
- [x] Edit row / card
- [x] Delete row(s)/ card

#### Checklist
- [x] Code reviewed

### DIFF
--- a/packages/data-explorer/src/components/dataView/DefaultCardContent.vue
+++ b/packages/data-explorer/src/components/dataView/DefaultCardContent.vue
@@ -1,6 +1,12 @@
 <template>
   <div>
     <h5 class="card-title mg-default-card-title">{{dataLabel}}
+      <a
+        class="btn btn-sm btn-link ml-1"
+        role="button"
+        :href="'/plugin/data-row-edit/' + dataTable + '/' + dataId">
+        <font-awesome-icon icon="edit"></font-awesome-icon>
+      </a>
       <slot name="shopping-button"></slot>
     </h5>
     <div class="card-text">
@@ -30,10 +36,10 @@
 <script>
 import Vue from 'vue'
 import { library } from '@fortawesome/fontawesome-svg-core'
-import { faSearch, faChevronUp, faChevronRight } from '@fortawesome/free-solid-svg-icons'
+import { faSearch, faChevronUp, faChevronRight, faEdit } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 
-library.add(faSearch, faChevronRight, faChevronUp)
+library.add(faSearch, faChevronRight, faChevronUp, faEdit)
 
 export default Vue.extend({
   name: 'DefaultCardContent',
--- a/packages/data-explorer/src/components/dataView/DefaultCardContent.vue
+++ b/packages/data-explorer/src/components/dataView/DefaultCardContent.vue
@@ -7,6 +7,13 @@
         :href="'/plugin/data-row-edit/' + dataTable + '/' + dataId">
         <font-awesome-icon icon="edit"></font-awesome-icon>
       </a>
+      <button
+        class="btn btn-sm btn-link"
+        role="button"
+        @click="$eventBus.$emit('delete-item', dataId)"
+        >
+        <font-awesome-icon icon="trash"></font-awesome-icon>
+      </button>
       <slot name="shopping-button"></slot>
     </h5>
     <div class="card-text">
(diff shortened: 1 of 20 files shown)


packages/data-explorer/src/components/dataView/DefaultCardContent.vue
```diff
@@ -1,6 +1,19 @@
 <template>
   <div>
     <h5 class="card-title mg-default-card-title">{{dataLabel}}
+      <a
+        class="btn btn-sm btn-link ml-1"
+        role="button"
+        :href="'/plugin/data-row-edit/' + dataTable + '/' + dataId">
```
MaxPostema, line 7: Add "plugin" to devServer proxy to be able to test the changes locally

packages/data-explorer/src/components/dataView/TableHeader.vue
```diff
@@ -2,6 +2,7 @@
   <thead>
   <tr>
     <th scope="col" v-if="isShop"></th>
+    <th scope="col" v-if="!isShop"></th>
```
MaxPostema, line 5: I don't see the reason to do this.. Is this useful?

packages/data-explorer/src/components/dataView/TableRow.vue
```diff
@@ -1,19 +1,45 @@
 <template>
   <tr>
     <td v-if="isShop"><shopping-button :isSelected="isSelected" :id="id"></shopping-button></td>
+    <td v-if="!isShop">
```
MaxPostema, line 4: In the default card view i can do shopping AND crud operations. This makes the table only have one or the other. 
I argue to make in consistent. 
Especially if we (later) add options to for example mass remove items via "shopping" selections. ( then you would like to be able to add to card in tableview. )

packages/data-explorer/src/components/dataView/TableRow.vue
```diff
@@ -1,19 +1,45 @@
...
+      <div class="btn-group" role="group" aria-label="row actions">
+        <button
+          class="btn btn-sm btn-link"
+          role="button"
+          @click="$eventBus.$emit('delete-item', id)"
```
MaxPostema, line 9: Nice trick. :)

packages/data-explorer/src/repository/dataRepository.ts
```diff
@@ -140,8 +140,13 @@ const getRowDataWithReferenceLabels = async (tableId: string, rowId: string, met
   return levelOneRowMapper(response.data, metaData)
 }
 
+const deleteRow = async (tableId: string, rowId: string) => {
+  return axios.delete(`/api/data/${tableId}/${rowId}`)
```
MaxPostema, line 144: Should there be state handeling if delete fails?... maybe the user is not allowed to delete this row and should be notified?

connoratrug: yes , should be passed to the action and handled there imo

packages/data-explorer/src/components/dataView/TableRow.vue
connoratrug: tnks, only one of the possible options, suggest to discuss these during the next meeting about the design

packages/data-explorer/src/components/dataView/TableRow.vue
connoratrug: i'm sorry i don't understand what you are saying here, do you suggest to add crud operations to the cart view ?, i would expect there to be replaced with cart actions ( remove from cart )

packages/data-explorer/src/components/dataView/TableHeader.vue
connoratrug: the column is there for two different reasons ( shop stuff or actions stuff), i agree the code looks silly now , i wanted to make it explicit so we can refactor this once we are clear on the UI/UX design

packages/data-explorer/src/components/dataView/DefaultCardContent.vue
connoratrug: o cool does work ? i will try it , good suggestion

connoratrug:
> There are some feedback items and a merge conflict.

merged with master and resolved conflict 

> ( ps: It was a bit annoying to have my state disappear after every crud operation. I hope that storing state in URL will improve the flow.)

Yes it will, however add and edit actions as separate app where designed to combine react bs-3 explorer with vue bs-4 app, to improve ux we could/should move add and edit actions into same app ( separate vue of modal) and use vue routing to switch view instead of page load.

PR #284, opened 2020-04-15
feat: Add greater and lesser rsql mapping for rangefilter and version bump

MaxPostema:
#### Checklist
- [x] Code reviewed

### DIFF
--- a/packages/data-explorer/src/mappers/rsqlMapper.ts
+++ b/packages/data-explorer/src/mappers/rsqlMapper.ts
@@ -10,8 +10,8 @@ export const createLikeQuery = (attributeName: string, selection: Value): Constr
  */
 export const createInQuery = (attributeName: string, selection: Value[]): Constraint => ({ selector: attributeName, comparison: ComparisonOperator.In, arguments: selection })
 export const createEqualsQuery = (attributeName: string, selection: Value): Constraint => ({ selector: attributeName, comparison: ComparisonOperator.Equals, arguments: selection })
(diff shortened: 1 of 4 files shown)


packages/data-explorer/src/mappers/rsqlMapper.ts
```diff
@@ -63,7 +66,14 @@ export const createRSQLQuery = (filters: FilterGroup, searchText?: string): stri
         operands.push(createLikeQuery(name, selection))
         break
       case 'range-filter':
-        operands.push(createRangeQuery(name, selection))
+        if (selection[0] == null && selection[1] !== null) {
+          operands.push(createLesserEqualQuery(name, selection[1]))
```
jelmerveen, line 70: i think you can write it like this: 

```case 'range-filter':
        if (selection[0] !== null && selection[1] !== null) {
          let correctedSelection = [Math.min(selection[0], selection[1]), Math.max(selection[0], selection[1])]
          operands.push(createRangeQuery(name, correctedSelection))
        } else {
          operands.push(createLesserEqualQuery(name, selection[0] || selection[1]))
        }
        break
```
to make it less complex

jelmerveen: and ofcourse the other conditions:

```
case 'range-filter':
        if (selection[0] !== null && selection[1] !== null) {
          let correctedSelection = [Math.min(selection[0], selection[1]), Math.max(selection[0], selection[1])]
          operands.push(createRangeQuery(name, correctedSelection))
        } else {
          operands.push(createLesserEqualQuery(name, selection[0] || selection[1]))
        }
        break
```

PR #289, opened 2020-05-04
feat: select table to explore

connoratrug:
- prepare: clean up ui to make space
  - move actions to toolbar
  - move headers stuff into header
  - move filters below header

#### Checklist
- [x] Code reviewed
-  User documentation updated
-  Updated javascript typing


packages/data-explorer/src/components/TableSelect.vue
```diff
@@ -0,0 +1,77 @@
...
+        toggle-class="bread-crumb-item-dropdown"
+        dropup
+        no-caret
+    >
+    <template v-slot:button-content>
+        <span class="bread-crumb-select">
+        <h1>
+            {{label}}
+        </h1>
+        <svg data-v-041edcbe=""
```
MaxPostema, line 15: Why not use a FontAwesomeIcon like used in the rest of the application?

packages/data-explorer/src/components/TableSelect.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+    <b-nav class="bread-crumb-dropdown">
+    <b-nav-item-dropdown
```
MaxPostema, line 3: Consider using some indentation here?

packages/data-explorer/src/components/TableSelect.vue
```diff
@@ -0,0 +1,77 @@
...
+            data-v-041edcbe=""
+            fill="currentColor"
+            d="M0 384.662V127.338c0-17.818 21.543-26.741 34.142-14.142l128.662 128.662c7.81 7.81 7.81 20.474 0 28.284L34.142 398.804C21.543 411.404 0 402.48 0 384.662z"
+            class="">
+            </path>
+        </svg>
+        </span>
+        </template>
+    <template v-for="table in packageTables">
+        <b-dropdown-item :key="table.id" :to="table.id">{{table.label}}</b-dropdown-item>
```
MaxPostema, line 34: Maybe add a close button to keep the style inline with the add/change filter box? I close boxes like this by clicking outside it. But having a close button might help other users, and makes it look complete.

packages/data-explorer/tests/unit/views/PageHeaderView.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  })
+
+  it('exists', () => {
+    const wrapper = shallowMount(PageHeaderView, { store, localVue })
+    expect(wrapper.exists()).toBeTruthy()
+  })
+
+  it('show table name and description', () => {
+    const wrapper = shallowMount(PageHeaderView, { store, localVue })
+    // todo replace with breadcrumb entity picker
```
MaxPostema, line 58: Are these comments relevant?

packages/data-explorer/src/components/TableSelect.vue
MaxPostema: What happens if there are 100 entities to switch to?

packages/data-explorer/tests/unit/views/PageHeaderView.spec.ts
connoratrug: yes, but will remove for now as breadcrumb can be build as part for new feature

packages/data-explorer/src/components/TableSelect.vue
connoratrug: this is the bootstrap ui default, suggest to keep it like this for now

connoratrug: > What happens if there are 100 entities to switch to?

a long list is shown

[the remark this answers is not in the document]
connoratrug: "Maybe a thin light gray line below it? Or always showing the caret?"

Both effect a ability to scan the page, switching i as secondary function, suggest to keep the ui clear, maybe mock up some suggestions first ?

packages/data-explorer/src/components/TableSelect.vue
connoratrug: because it differs from the chevron used in bootstrap ( look at menu and filters )

packages/data-explorer/src/components/TableSelect.vue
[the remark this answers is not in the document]
connoratrug: The idea is to at some point have a breadcrumb trail to the package root, brumbs could be separated to right-chevron, suggest to keep it clean for now, most non regular users do not need to use this feature

PR #296, opened 2020-05-12
feat: add a data display limit and warning

MaxPostema:
#### Checklist
- [ ] Code reviewed


packages/data-explorer/src/repository/dataRepository.ts
```diff
@@ -93,7 +93,7 @@ const addFilterIfSet = (request: string, rsqlFilter?: string): string => {
   return rsqlFilter ? `${request}&q=${encodeRsqlValue(rsqlFilter)}` : request
 }
 
-const getTableDataDeepReference = async (tableId: string, metaData: MetaData, coloms: string[], rsqlQuery?: string) => {
+const getTableDataDeepReference = async (tableId: string, metaData: MetaData, coloms: string[], size: Number, rsqlQuery?: string) => {
```
connoratrug, line 96: Would suggest to make size optional, and maybe use met more descriptive term or add js doc describing the params

packages/data-explorer/src/components/utils/EndOfResults.vue
```diff
@@ -0,0 +1,19 @@
...
+
+export default Vue.extend({
+  name: 'EndOfResults',
+  props: {
+    dataDisplayLimit: {
+      type: Number,
+      required: true
+    }
+  }
+})
```
connoratrug, line 18: suggest to move to domain specific folder ( for example , dataview or something like that) instead of generic 'util' folder. If it is truly a global component in the context of the app would suggest to just place it in the components folder.

PR #298, opened 2020-05-12
feat: Filters version bump and add compounds labels to filter list

MaxPostema:
Note: found bug in api doc's metadata v3. 'parentAttribute' needed to be 'parentAttributeId'

#### Checklist
- [ ] Code reviewed


packages/data-explorer/src/mappers/metaFilterMapper.ts
```diff
@@ -21,11 +21,15 @@ const fieldTypeToFilterType:StringMap = {
...
   'onetomany': 'multi-filter',
   'enum': 'checkbox-filter',
-  'file': 'string-filter'
+  'file': 'string-filter',
+  'compound': 'compound-title'
 }
 
 const mapMetaToFilters = async (metaData: MetaData) => {
   let shownFilters:string[] = []
+  function findByID (ID: string):Attribute {
```
connoratrug, line 30: why not id , instead of ID

packages/data-explorer/src/mappers/metaFilterMapper.ts
```diff
@@ -21,11 +21,15 @@ const fieldTypeToFilterType:StringMap = {
...
+    return metaData.attributes.filter((item) => item.id === ID)[0]
```
connoratrug, line 31: what if filter returns empty list ?

PR #306, opened 2020-06-11
fix: redirect to login page on 401

jvanveen:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/data-explorer/src/lib/api.ts
+++ b/packages/data-explorer/src/lib/api.ts
@@ -1,0 +1,14 @@
+import axios from 'axios'
+
+const api = axios.create({
+  baseURL: '/',
+  timeout: 1000,
+  validateStatus: function (status) {
+    if (status === 401) {
+      window.location.href = '/login'
+    }
+    return status >= 200 && status < 300
+  }
+})
+
+export default api
--- a/packages/data-explorer/src/repository/dataRepository.ts
+++ b/packages/data-explorer/src/repository/dataRepository.ts
@@ -3,7 +3,7 @@ import { getAttributesfromMeta } from './metaDataService'
 import * as metaDataRepository from './metaDataRepository'
 import { DataApiResponse, DataApiResponseItem, DataObject } from '@/types/ApiResponse'
 import { MetaData, Attribute } from '@/types/MetaData'
-import axios from 'axios'
+import api from '@/lib/api'
 import { encodeRsqlValue } from '@molgenis/rsql'
 
 // maps api response to object with as key the name of the column and as value the label of the value or a list of labels for mrefs
@@ -115,7 +115,7 @@ const getTableDataDeepReference = async (
   const expandReferencesQuery = buildExpandedAttributesQuery(metaData, coloms)
   const size = _buildSizeQueryParam(dataDisplayLimit)
   const request = addFilterIfSet(`/api/data/${tableId}?${size}${expandReferencesQuery}`, rsqlQuery)
-  const response = (await axios.get<DataApiResponse>(request)).data
+  const response = (await api.get<DataApiResponse>(request)).data
   const result = { items: response.items.map((item: DataApiResponseItem) => levelNRowMapper(item)) }
   return result
 }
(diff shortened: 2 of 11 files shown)


packages/data-explorer/src/views/PageHeaderView.vue
```diff
@@ -63,7 +63,9 @@ export default Vue.extend({
     ])
   },
   mounted () {
-    this.getGroupTabels({ package: this.tableMeta.package })
+    if (this.tableMeta) {
```
connoratrug, line 66: the if should not be needed and would break the flow, the tableMeta should be there on mount

packages/data-explorer/src/lib/api.ts
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios'
+
+const apiClient = axios.create({
+  baseURL: '/',
+  timeout: 1000,
+  validateStatus: function (status) {
+    if (status === 401) {
+      window.location.href = '/login'
+    }
+    return true
```
connoratrug, line 10: What happens to the original validateStatus code ? 
`return status >= 200 && status < 300; // defaull`
I think we would need both

packages/data-explorer/src/repository/dataRepository.ts
```diff
@@ -130,7 +130,7 @@ const getTableDataWithLabel = async (tableId: string, metaData: MetaData, column
   const expandReferencesQuery = buildExpandedAttributesQuery(metaData, [...columnSet])
   const size = _buildSizeQueryParam(dataDisplayLimit)
   const request = addFilterIfSet(`/api/data/${tableId}?${size}${expandReferencesQuery}`, rsqlQuery)
-  const response = (await axios.get<DataApiResponse>(request)).data
+  const response = (await api.get<DataApiResponse>(request)).data
```
connoratrug, line 133: i like we get to keep the Typed request 👍

packages/data-explorer/src/lib/api.ts
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios'
+
+const apiClient = axios.create({
```
connoratrug, line 3: Maybe would could use 'transformResponse' to handle errors, could also to the 'login' redirect'

packages/data-explorer/src/lib/api.ts
```diff
@@ -0,0 +1,14 @@
...
+  baseURL: '/',
+  timeout: 1000,
+  validateStatus: function (status) {
```
connoratrug, line 6: possibly use `axios.interceptors` instead ? lets discuss

packages/data-explorer/src/lib/api.ts
jvanveen: validateStatus lijkt hiervoor bedoeld

packages/data-explorer/src/lib/api.ts
jvanveen: Zie comment hierboven

packages/data-explorer/src/views/PageHeaderView.vue
jvanveen: Removed

packages/data-explorer/src/lib/api.ts
jvanveen: Added default behaviour back in

packages/data-explorer/src/lib/api.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (status === 401) {
```
connoratrug, line 7: what about 404 ?

packages/data-explorer/src/lib/api.ts
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios'
```
connoratrug, line 1: maybe client or something like that instead op api ? , api is pretty generic

packages/data-explorer/src/lib/api.ts
jvanveen: Volgens mij is dat puur een configuratie fout geweest in mijn lokale Molgenis instance. Een 401 is de correcte code. Op basis van een 404 zou je volgens mij niet de aanname mogen maken dat er ingelogd moet worden. Anders moeten we het misschien op een andere manier oplossen.

connoratrug:
suggest we add a unit test for the client

PR #307, opened 2020-06-12
fix: refactored unittests

jelmerveen:
#### Checklist
- [x] Code reviewed
-  User documentation updated
-  Updated javascript typing

### DIFF
--- a/packages/data-explorer/tests/unit/store/mutations.spec.ts
+++ b/packages/data-explorer/tests/unit/store/mutations.spec.ts
@@ -1,4 +1,5 @@
 import mutations from '@/store/mutations'
+import mockState from '../mocks/mockState'
 import ApplicationState from '@/types/ApplicationState'
 import { Attribute, MetaData } from '@/types/MetaData'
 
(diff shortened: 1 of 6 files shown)


connoratrug:
it's not really a fix as no issue is fixed, you can use the `task:` commit fag to signal no version bump should be made

jelmerveen:
> it's not really a fix as no issue is fixed, you can use the `task:` commit fag to signal no version bump should be made

we'll fix that with a squash! Task is not a commitizen, but chore / refactor are. We'll use refactor.

packages/data-explorer/tests/unit/store/mutations.spec.ts
```diff
@@ -51,35 +52,8 @@ describe('mutations', () => {
...
-        hideSidebar: false,
-        definition: [],
-        shown: [],
-        selections: {}
-      },
-      searchText: '',
-      bookmark: '',
-      componentRoute: false
-    }
+    baseAppState = mockState()
```
connoratrug, line 55: Personally i am not a fan of importing a general mock into multiple test, as the creates interdependencies that would otherwise not be needed. In addition mocking more than needed for the test makes is harder to read and debug the test. The need for this kind of mock signals a dependency that probably needs to be refactored to be less interdependent . As a alternative you could create a mockstate in de test `state:any = { my-prop: 'needed-in-test'}`.

PR #353, opened 2020-08-27
Feat/bread crumb bar

connoratrug:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/data-explorer/src/store/header/actions.ts
+++ b/packages/data-explorer/src/store/header/actions.ts
@@ -2,7 +2,7 @@ import client from '@/lib/client'
 import { DataApiResponse, isDataApiResponseItem } from '@/types/ApiResponse'
 import { Breadcrumb } from '@/types/BreadCrumb'
 
-const getBreadCrumbPath = async (parentUrl: string, addBreadcrumb: Function, buildLink: Function): Promise<any> => {
+const getBreadcrumbPath = async (parentUrl: string, addBreadcrumb: Function, buildLink: Function): Promise<any> => {
   const packageName = parentUrl.split('/').pop()
   const query = `q=id==${packageName}`
   const resp = await client.get<DataApiResponse>(`/api/data/sys_md_Package?${query}`)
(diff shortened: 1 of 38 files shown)


packages/data-explorer/src/store/header/actions.ts
```diff
@@ -1,5 +1,28 @@
 import client from '@/lib/client'
-import { DataApiResponse } from '@/types/ApiResponse'
+import { DataApiResponse, isDataApiResponseItem } from '@/types/ApiResponse'
+
+const getBreadCrumbPath = async (parentUrl: string, addBreadcrumb: Function, buildLink: Function): Promise<any> => {
```
jelmerveen, line 4: To keep it consistent: getBreadcrumbPath

packages/data-explorer/src/store/header/actions.ts
```diff
@@ -16,5 +39,24 @@ export default {
     })
     commit('setPackageTables', packageTables)
     return packageTables
+  },
+
+  fetchBreadcrumbs: async ({ commit, getters, rootState }: { commit: any, getters: any, rootState: any }) => {
```
jelmerveen, line 44: I would recommend handling the correct order here, else you have to have in (your own) memory that the order is reversed.

Seeing mutations (unshift, instead of push) and  the breadcrumbbar.vue depends on a reverse array

If you fetch it, and then reverse the array one time, then there is upstream no more troubles.

jelmerveen (changes requested):
Please see my comment

packages/data-explorer/src/store/header/actions.ts
connoratrug: done

packages/data-explorer/src/store/header/actions.ts
connoratrug: still feels a little strange to have order different from other on screen, but i flipped it

jelmerveen (approved):
Nice solution doing it in the component, which actually makes more sense.

PR #379, opened 2020-09-23
Task/add settings ADR

connoratrug:
#### Checklist
- [ ] Code reviewed


doc/adr/0003-have-app-create-settings.md
```diff
@@ -0,0 +1,26 @@
...
+
+Apps (for example the data-explorer-2 or the biobank-explorer) may need settings data to function properly. As these settings are dynamic (may be changed at run time), and may change between deployments these settings need to be stored somewhere (molgenis database). This poses two problems; 1) how to link the app with the settings data, 2) how to control access to the settings.
+
+## Decision
+
+We leave it up to the app to create the settings entity. Each app creates 0 or 1 settings entities. By default everyone can read the settings data. Depending on the app the entity contains 1 or more rows ( in case of multiple rows the app is responsible using the correct row(s) in each context).
+
+Rejected alternatives:
+
+- Add settings via bootstrapped entity in molgenis core; This was deemed as insufficiently flexible due to the dependance on Java knowledge and (possible major version) core release.
```
mswertz, line 19: did you consider the json blob method?

connoratrug: @mswertz it depends what you mean by 'the blob method', could you elaborate ? We did discuss creating a new endpoint connected to some settings blob, this was rejected as this does not link up with the molgenis security model. Molgenis at its core uses entities and attributes. It therefore seems logical to use these instead of creating a new blob infrastructure for storing settings. 

"This was deemed as insufficiently" this would indeed be 'solved' my using a non typed data structure in an entity. The ADR does not prohibit ( nor enforces) the use of a (single)blob column.  What remains is the dependance on Java and dependance on a backend release. Although the ADR does not discuss this i personally would still recommend creating default settings tabels for the core apps ( data-explorer, ..) at molgenis bootstrap time. This was however perviously rejected by yourself, @fdlk and @tommydeboer as being insufficiently flexible ( see reasons mentioned in PR)

connoratrug:
[#7490 Apps settings](https://trello.com/c/WYAKoq7N/275-7490-apps-settings)

mswertz (approved):

PR #382, opened 2020-09-24
feat(components-library): added passthrough classes so you can style …

jelmerveen:
…the toast

#### Checklist
- [x] Code reviewed

### DIFF
--- a/packages/components-library/src/components/CartSelectionToast.vue
+++ b/packages/components-library/src/components/CartSelectionToast.vue
@@ -5,6 +5,9 @@
     no-auto-hide
     :toaster="location"
     :variant="backgroundVariant"
+    :toastClass="toastClass"
+    :bodyClass="bodyClass"
+    :headerClass="headerClass"
   >
     <div class="container">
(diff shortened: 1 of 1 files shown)


MaxPostema (approved):

packages/components-library/src/components/CartSelectionToast.vue
```diff
@@ -5,6 +5,9 @@
     no-auto-hide
     :toaster="location"
     :variant="backgroundVariant"
+    :toastClass="toastClass"
```
connoratrug, line 8: :S really don't like the way this scales, new prop for every class detail , maybe use a slot instead ?

connoratrug:
[Customizability cart selection toast](https://trello.com/c/Qt34S0RS/278-customizability-cart-selection-toast)

packages/components-library/src/components/CartSelectionToast.vue
jelmerveen: This something that a b-toast has, I only expose it in our wrapper. nothing more nothing less. Has nothing to do with scalability

PR #385, opened 2020-09-24
feat(components-library): Add string-filter component

connoratrug:
- Move String filter over from filter repo to component repo
- Extend test to cover missing cases
- Fix lint issues
- Add StringFilter to build uitput

#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/components-library/src/components/filters/StringFilter.vue
+++ b/packages/components-library/src/components/filters/StringFilter.vue
@@ -1,0 +1,77 @@
+<template>
+  <b-input-group>
+    <b-form-input
+      v-model="model"
+      :name="name"
+      :placeholder="placeholder"
+      trim
+    />
+    <b-input-group-append>
+      <b-button
+        variant="outline-secondary"
+        @click.prevent="model=''"
+      >
+        <font-awesome-icon icon="times" />
+      </b-button>
+    </b-input-group-append>
+  </b-input-group>
+</template>
+
+<script>
+import Vue from 'vue'
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { faTimes } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+library.add(faTimes)
+export default Vue.extend({
+  name: 'StringFilter',
+  components: { FontAwesomeIcon },
+  props: {
+    /**
+     * The HTML input element name.
+     */
+    name: {
+      type: String,
+      required: true
+    },
+    /**
+     * The HTML input element placeholder.
+     */
+    placeholder: {
+      type: String,
+      required: false,
+      default: () => ''
+    },
+    /**
+     * The StringFilter model value.
+     * @model
+     */
+    value: {
+      type: String,
+      default: () => ''
+    }
+  },
+  computed: {
+    model: {
+      get () {
+        return this.value || ''
+      },
+      set (value) {
+        this.$emit('input', value === '' ? undefined : value)
+      }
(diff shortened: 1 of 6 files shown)


connoratrug:
[Add String filter to component lib](https://trello.com/c/QXVIuYs4/279-add-string-filter-to-component-lib)

packages/components-library/src/components/filters/StringFilter.vue
```diff
@@ -0,0 +1,77 @@
...
+      required: false,
+      default: () => ''
+    },
+    /**
+     * The StringFilter model value.
+     * @model
+     */
+    value: {
+      type: String,
+      default: () => ''
```
jvanveen, line 51: Misschien deze ook required maken dan?

packages/components-library/tests/unit/filters/StringFilter.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  it('sets value property on inner input', () => {
+    const inputElement = wrapper.find('input').element as HTMLInputElement
+    expect(inputElement.value).toBe('value')
+  })
+
+  it('emits undefined when clear button is clicked', () => {
+    wrapper.find('button').trigger('click')
+    // @ts-ignore
```
jvanveen, line 32: Zijn die ts-ignore statements nog nodig?

packages/components-library/tests/unit/filters/StringFilter.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  describe('not passing a model prop', () => {
+    const wrapper = mount(StringFilter, {
+      localVue,
+      propsData: {
+        name: 'name',
+        label: 'label'
+      }
+    })
```
jvanveen, line 52: Hier een getWrapper van maken, net zoals CheckboxFilter.spec.ts?

packages/components-library/tests/unit/filters/StringFilter.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  beforeEach(() => {
+    wrapper = mount(StringFilter, {
+      localVue,
+      propsData: {
+        name: 'name',
+        label: 'label',
+        value: 'value'
+      }
+    })
+  })
```
jvanveen, line 19: Hier een getWrapper van maken, net zoals CheckboxFilter.spec.ts?

jvanveen (approved):
Ziet er goed uit. Ship it wmb, tenzij je nog wat wil doen met die suggesties.

packages/components-library/src/components/filters/StringFilter.vue
connoratrug: Suggest to keep it , goal is to move over the components from one repo to other repo instead of updating / trimming the components

packages/components-library/tests/unit/filters/StringFilter.spec.ts
connoratrug: yes it is , the emitted function is not guaranteed to have a input prop, but for the test i do not care about this warning ( in app code i would )

packages/components-library/tests/unit/filters/StringFilter.spec.ts
connoratrug: current solution seems more clear / simpel , agree that is complex setup is need a helper function is nice, in general unit test should try to avoid complex setups imo , if complex setups are needed this to me signals that the unit under test might not we very modulair.

PR #392, opened 2020-09-30
Feat/add change filter comp

connoratrug:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/components-library/src/components/filters/ChangeFilters.vue
+++ b/packages/components-library/src/components/filters/ChangeFilters.vue
@@ -75,7 +75,7 @@ export default Vue.extend({
      */
     value: {
       type: Array,
-      default: () => []
+      required: true
     }
   },
   data () {
--- a/packages/components-library/src/components/filters/ChangeFilters.vue
+++ b/packages/components-library/src/components/filters/ChangeFilters.vue
@@ -1,0 +1,193 @@
+<template>
+  <div>
+    <b-dropdown
+      ref="addFilter"
+      variant="primary"
+      boundary="window"
+      menu-class="shadow ml-2"
+      dropright
+      no-caret
+      block
+    >
+      <template v-slot:button-content>
+        Change filters
+        <font-awesome-icon icon="caret-right" class="ml-2" />
+      </template>
+      <b-dropdown-text>
+        Change filters
+        <span
+          class="float-right remove-button"
+          @click.stop="$refs.addFilter.hide(true)"
+        >
+          <font-awesome-icon icon="times" />
+        </span>
+      </b-dropdown-text>
+      <b-dropdown-form>
+        <b-form-checkbox
+          v-for="option in filters"
+          :key="option.name"
+          :checked="selected.includes(option.name)"
+          :class="{'ml-4':option.compound}"
+          @change="checkboxHandler(option, $event)"
+        >
+          <span v-if="option.label">
+            <span class="text-nowrap">{{ option.label }}</span>
+            <span class="text-secondary">
+              <small>
+                <span class="text-nowrap">( {{ option.name }} )</span>
+              </small>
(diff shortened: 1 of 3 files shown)


connoratrug:
[#7576 Add change filter component](https://trello.com/c/OUQ6C5UO/289-7576-add-change-filter-component)

packages/components-library/src/components/filters/ChangeFilters.vue
```diff
@@ -0,0 +1,200 @@
...
+### Usage
+```jsx
+const filters = [
+  {
+    name: 'color',
+    label: 'Color',
+  },
+  {
+    name: 'Region',
+    label: 'region',
```
jvanveen, line 182: Label should be upper-case instead of name?

packages/components-library/src/components/filters/ChangeFilters.vue
```diff
@@ -0,0 +1,200 @@
...
+.remove-button:hover {
+  color: var(--danger);
+}
+</style>
+
+<docs>
+Filter selector
+### Usage
+```jsx
+const filters = [
```
jvanveen, line 175: I'm missing some of the features that the test reveals, like compound. Maybe tweak the example filters data a bit to show the other options?

packages/components-library/tests/unit/filters/ChangeFilters.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  it('can add a filter', (done) => {
+    wrapper.find('button.dropdown-toggle').trigger('click')
+    wrapper.vm.$nextTick(() => {
+      wrapper.findAll('.custom-control-input').at(1).trigger('click')
+      wrapper.vm.$nextTick(() => {
+        expect(wrapper.emitted().input[0]).toEqual([['search', 'compound-title', 'string', 'checkbox']])
+        done()
+      })
+    })
+  })
```
jvanveen, line 53: Converting all tests to async/await?

packages/components-library/tests/unit/filters/ChangeFilters.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+          })
+        })
+      })
+    })
+  })
+
+  it('can make a array unique', () => {
+    // @ts-ignore
+    expect(['a', 'b', 'b', 'c', 'c', 'c'].filter(wrapper.vm.unique)).toEqual(['a', 'b', 'c'])
+  })
```
jvanveen, line 111: Rather than testing internal component functions, wouldn't it be better to check the rendered output for double values where you wouldn't expect it? From vue-test-utils:
```
Instead, we recommend writing tests that assert your component's public interface, and treat its internals as a black box.
 ```

jvanveen (changes requested):

packages/components-library/src/components/filters/ChangeFilters.vue
connoratrug: updated case

packages/components-library/src/components/filters/ChangeFilters.vue
connoratrug: added compound to the example

packages/components-library/tests/unit/filters/ChangeFilters.spec.ts
connoratrug: done

packages/components-library/tests/unit/filters/ChangeFilters.spec.ts
connoratrug: yes, why is unique function part of this component ?

jvanveen (approved):
Looks good to me. Ship it

PR #397, opened 2020-10-07
chore: #7309 add molgenis core apps adr

jvanveen:
#### Status
- [x] Approval
- [ ] Rejected
- [x] Accepted

# Seen
- [ ] Connor
- [ ] Luuk
- [x] Jelmer
- [ ] Jeroen
- [x] Max
- [x] Morris

### DIFF
--- a/doc/adr/0004-molgenis-core-apps.md
+++ b/doc/adr/0004-molgenis-core-apps.md
@@ -1,0 +1,55 @@
+# 4. Molgenis Core Apps
+
+Date: 2020-09-23
+
+## Status
+
+Proposal
+
+## Context
+
+See [frontend architecture sessions](https://docs.google.com/document/d/1VW3ah5VAvAz2KnqNZlNmVqCzFhBMlIcjPPUlsHMFRIY/) for additional background
+information on micro-frontend/SPA architecture.
+
+Molgenis is flexible software that allows users to create their own apps.
+This is quite powerful whenever a new, custom application needs to be made
+on top of the Molgenis backend. However, the same modular paradigm is also
+being applied to parts of the frontend, that are always being shipped with
+Molgenis.
+
+At this time, we have approximately 12 legacy Molgenis apps that are still
+rooted in Molgenis freemarker templates and Java controllers. There are
+about 11 new-style Unpkg frontend applications. The Unpkg applications
+all come with their own package configuration, release cycle, build tools
+and testsuites; all glued together in the Nginx proxy, so they are
+approachable from the same domain.
+
+## Decision
+
+The keyword here is convergence; the act of converging and especially moving
+toward union or uniformity. This means that Molgenis frontend core is going
+to be one Vue application, in which all components from the 11 existing Unpkg
+applications are merged together in a single frontend package. Instead of
+glueing them together with Nginx, we will do this using Vue-router and
+Vuex modules. Instead of having 11 app versions, we will use one frontend
+version and use git tags to mirror each Molgenis release.
+
+## Consequences
+
+This [frontend spike](https://github.com/molgenis/frontend-spike) already
+proved that merging existing Unpkg apps together isn't that technically
+challenging. Vuex modules provides namespacing with minimal changes.
+
+An open question is how to integrate legacy applications in the resulting SPA.
+The answer is: we shouldn't attempt to integrate them to close.  I've seen several attempts so far to integrate legacy applications in a SPA, using iframe wrappers
+and IPC.
+
+Instead of spending energy and time into such a workaround, we should just keep
+the styling and layout of both solutions in sync, until the last legacy
+module has been moved to the new frontend. We should still use the same
+unpkg mechanism, but we just point it to the new frontend package using a
+URL like: https://master.dev.molgenis.org/v2/
+
+The Nginx proxy config needs to be updated, to be able to support history
+routing on this endpoint, so that uri's like **/v2/dataexplorer** and **/v2/navigator**
+route to the correct frontend view.
--- a/doc/adr/0004-molgenis-core-apps.md
+++ b/doc/adr/0004-molgenis-core-apps.md
@@ -8,26 +8,27 @@ Proposal
 
 ## Context
 
-See [frontend architecture sessions](https://docs.google.com/document/d/1VW3ah5VAvAz2KnqNZlNmVqCzFhBMlIcjPPUlsHMFRIY/)
-for additional background information on micro-frontend/SPA architecture.
+*See [frontend architecture sessions](https://docs.google.com/document/d/1VW3ah5VAvAz2KnqNZlNmVqCzFhBMlIcjPPUlsHMFRIY/)
+for background information on micro-frontend/SPA architecture.*
 
-Molgenis is flexible software that enables its users to create their own apps.
-This is useful for custom applications, made on top of the Molgenis backend.
-However, the same micro-architecture paradigm is also applied to parts of
-the frontend that are *always* being shipped with Molgenis. We call these
-admin-like Molgenis apps "core packages".
+Molgenis is flexible software that enables its users to create their own
+frontend apps. This is useful for custom applications, made on top of a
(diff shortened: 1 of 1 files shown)


jvanveen:
[#7309 ADR: Molgenis Core Apps](https://trello.com/c/qRSkAklk/298-7309-adr-molgenis-core-apps)

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,55 @@
...
+about 11 new-style Unpkg frontend applications. The Unpkg applications
+all come with their own package configuration, release cycle, build tools
+and testsuites; all glued together in the Nginx proxy, so they are
+approachable from the same domain.
+
+## Decision
+
+The keyword here is convergence; the act of converging and especially moving
+toward union or uniformity. This means that Molgenis frontend core is going
+to be one Vue application, in which all components from the 11 existing Unpkg
```
jelmerveen, line 31: The big question(s) is/are: Do we need to be able to update the parts seperately and are some of the features 'opt-in' ?

jelmerveen: Would you consider using a multi-page setup to combine the apps together, instead of one big merge?

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,55 @@
...
+applications are merged together in a single frontend package. Instead of
+glueing them together with Nginx, we will do this using Vue-router and
+Vuex modules. Instead of having 11 app versions, we will use one frontend
+version and use git tags to mirror each Molgenis release.
```
jelmerveen, line 35: if we have one big spa, then we need to build everything everytime, with the current E2E tests this would lead to build times of over 30 minutes for these tests alone

MaxPostema:
I don't think it is necessary to serve the legacy molgenis apps via SPA? communication with iframes may still create complex situations. May be better to keep legacy as separate and lose SPA context on loading them. ( nice reasons to update the legacy apps )

doc/adr/0004-molgenis-core-apps.md
jvanveen: That would be an interesting option if we decide to merge all buildtools from all packages together, while still keeping multiple apps around. However, I think we could make better use of Vue-router and Vuex by having only one application with multiple views. The main reasons:
* We could apply history routing to all parts of the application, making the application as a whole feel much more responsive.
* Easier dependency management, since we only have to include dependencies once. While we keep tests and sandbox environments separate, we still have to include all dependencies for each package(as dev-dependencies, but still...)

We could start a migration path by implementing just one simple app in a new package (e.g. core-app); for example a theme selector. The time we spend there would be useful to decide on definitive tooling. After this, we could migrate an existing package into the core-app; for example DE2. 

Right now, the current build toolchain (Webpack/Babel) is a bottleneck. As soon we move in more components in the build, we would start to see unacceptable buildtimes during development or while building for production(less urgent). More recent toolchains like Vite and Snowpack rely on ES modules during development and a high performance bundler (esbuild). That would definitly fix slow build times during development.

Webpack could use esbuild, but tweaking Webpack's config is complex, even more with the Vue-CLI wrapper around it. Vite(https://github.com/vitejs/vite#how-and-why) seems an interesting alternative. It uses the same kind of config as vue-cli, but without Webpack. Sadly, it has no Vue 2 support and Bootstrap-vue doesn't have a Vue 3 version yet(https://github.com/bootstrap-vue/bootstrap-vue/issues/5196). So that limits our Vite options to either waiting for a Vue3 version of Bootstrap-Vue and stick with Vue2 for now, or dropping Bootstrap-Vue and migrate each app to Vue 3.

Another option would be to use Snowpack together with Vue 2. I've tried Snowpack a while and it works really well. However, I think we should definitely give Vite a try first, since it treats Vue as a first-class citizen.

doc/adr/0004-molgenis-core-apps.md
jvanveen: While on this migration path, I think we should go through the E2E testsuite and rewrite what's still useful to the new Playwright setup. E2E tests on a SPA with a tool like Playwright are going to be much faster to execute. However, it could still be that the backend forms a bottleneck. We could either optimize a backend for our E2E tests(e.g. https://playwright.test.molgenis.org/) or defer E2E tests to a once-a-day test, instead of checking on each merge to master.

jvanveen:
> I don't think it is necessary to serve the legacy molgenis apps via SPA? communication with iframes may still create complex situations. May be better to keep legacy as separate and lose SPA context on loading them. ( nice reasons to update the legacy apps )

Totally agree! An hybrid iframe solution tends to be complicated and expensive to maintain.

mswertz:
Hi. I support exploration if merging of navigator and explorer is beneficial. I also see drawbacks in terms of complexity. I also see potential advantages for reducing number of apps. However, I am not yet convinced that one 'core' app to rule them all is useful. So I don't yet support ADR into that direction.

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,88 @@
...
+juggle with, less toolchains to maintain, less packages to install and a faster
+user-experience in the app, because state between core apps is maintained
+as a SPA using Vue-router. Apps that use Vuex are being compartimentalized
+within the core app, using Vuex modules.
+
+## Decision
+
+The keyword here is convergence; the act of converging and especially moving
+toward union or uniformity. This means that Molgenis core is going to be one
+Vue application. Conceptually, it is no different than the existing unpkg-style
```
mswertz, line 65: This decision I don't support yet. However, the decision to try and reduce number of apps and evaluate that that is beneficial I do support.

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,88 @@
...
+Molgenis is flexible software that allows users to create their own apps.
+This is quite powerful whenever a new, custom application needs to be made
+on top of the Molgenis backend. However, the same modular paradigm is also
+being applied to parts of the frontend that are always being shipped with
+Molgenis. We call these admin-like apps Molgenis "core" packages.
+
+As an assessment; there are at the moment approximately 11 legacy Molgenis apps
+that are still rooted in Molgenis freemarker templates and Java controllers:
+
+* feedback
```
jelmerveen, line 23: This data won't age well. Maybe you should describe it as:

Whenever we make a new application for a feature and we are going to use that in production, we will always look to consolidate this app within exisiting apps if any other apps already have similarities. Either by user role (like admin tools) or kinship

jvanveen: Sounds good.

doc/adr/0004-molgenis-core-apps.md
jvanveen: Updated

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,34 @@
...
+*See [frontend architecture sessions](https://docs.google.com/document/d/1VW3ah5VAvAz2KnqNZlNmVqCzFhBMlIcjPPUlsHMFRIY/)
+for background information on micro-frontend/SPA architecture.*
+
+## Decision
+
+We are going to merge the following apps together in a *@molgenis-ui/core* app:
+
+* navigator
+* data explorer 2
+* searchall
```
mswertz, line 20: verwijderen.

doc/adr/0004-molgenis-core-apps.md
```diff
@@ -0,0 +1,33 @@
...
+* navigator
+* data explorer 2
+* data-row-edit
+
+## Consequences
+
+* Pick the preferred index file delivery method (through separate ADR)
+* use Vue-router to connect the views together
+* separate Vuex stores with Vuex modules
+* Use clean history api urls, e.g.:
```
mswertz, line 27: bezwaar

mswertz (approved):

PR #440, opened 2020-11-19
feat: add download with toaster

jvanveen:
#### Checklist
- [ ] Code reviewed


connoratrug (changes requested):
- Add missing unit tests for new functions
- functional issues: 
   - When a download results in an error and i try to read the error, the toast disappears.  It should not.
   - When i try to down load the 'People' entity i fist see a small bue bar followed by an error, that then disappears

connoratrug (approved):

PR #450, opened 2020-12-07
fix(data-row-edit): recursively build reference map

connoratrug:
Needed when compound includes reference types
Traverse attribute tree to flatten attributes before filtering ref-types and building map

#### Checklist
- [x] Code reviewed
-  User documentation updated
-  No warnings during install
-  Updated javascript typing


connoratrug:
[Mref form component does not show add-option action when inside compound](https://trello.com/c/sRumPnn5/344-mref-form-component-does-not-show-add-option-action-when-inside-compound)

MaxPostema (approved):

PR #464, opened 2020-12-11
Feat/7740 pagination

jvanveen:
#### Checklist
- [ ] Code reviewed


packages/data-explorer/src/views/MainView.vue
```diff
@@ -143,24 +146,30 @@ export default Vue.extend({
...
       } else {
-        this.fetchTableViewData()
+        res = await this.fetchTableViewData(this.pagination)
       }
       this.loading = false
+      return res
+    },
+    async paginator () {
+      const res = await this.fetchViewData(this.$route.params.entity)
+      return { count: res.response.page.totalElements }
```
connoratrug, line 158: would suggest to move this into the store, then there would be no need to move around the 'res' on the outside

jvanveen: Moved fetchCardViewData to actions

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+    </div>
+
+    <form v-if="val.count > val.size" class="pagination-controls form-inline ml-2">
+      <div class="btn-group">
+        <select  class="form-control-sm ml-2 mr-2" v-model="val.size">
+          <option value="10">10</option>
+          <option value="20">20 </option>
+          <option value="50">50</option>
+        </select>
+        <label for="exampleInputEmail1">per page</label>
```
connoratrug, line 49: maybe expose via prop or use i18n to change / translate label

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+      </div>
+    </form>
+  </div>
+
+  <div class="item-count mt-2">{{val.count}} entities total</div>
```
connoratrug, line 54: maybe expose via prop or use i18n to change / translate label

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+</nav>
+</template>
+
+<script>
+export default {
```
connoratrug, line 59: missing 'name' attr in component , suggest using following attr order:
name, components, props, computed, methods, watch, lifecycle

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+        for (let i = this.val.page - midPage; i < this.val.page; i++) {
+          if (i > 0) { pages.push(i) }
+        }
+        // Generate the second part of the middle range.
+        for (let i = this.val.page; i <= this.val.page + midPage; i++) {
+          if (i > 0 && i <= this.pageCount) { pages.push(i) }
+        }
+      } else {
+        // Render until pageCount or relativePageSize max.
+        const _r = (this.pageCount > this.midPageSize) ? this.midPageSize : this.pageCount
```
connoratrug, line 102: what does _r stand for ? suggest using more descriptive name

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+      // The same with the end part:
+      const lastPage = pages[pages.length - 1]
+      if (lastPage === (this.pageCount - 1)) { pages.push(this.pageCount) }
+
+      return pages
+    }
+  },
+  data: function () {
+    return {
+      val: { size: 20, page: 1, loading: false, count: 0 }
```
connoratrug, line 121: suggest using more descriptive name , having 'val', and value seems confusing

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+    styles: {
+      type: Object,
+      required: false,
+      default: () => ({})
+    },
+    /**
+     * The method to retrieve items with.
+     * @returns {Object} {count: Number}
+     */
+    method: {
```
connoratrug, line 177: suggest using more descriptive name , ( maybe retrieveItems, or retrieveItemsFunction, or fetchItems, or getItems )

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+     */
+    useRouter: {
+      type: Boolean,
+      default: () => true
+    },
+    /**
+     * The internal pagination model
+     * @model
+     */
+    value: {
```
connoratrug, line 199: This is can not really be the internal model as it gets passed from outside and there is also the val field, is this the initial model ?

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+  },
+  watch: {
+    '$route.query': {
+      handler: function (query) {
+        if (!this.method || !this.useRouter) { return }
+
+        if (query.page) {
+          this.fetchData(parseInt(query.page, 10))
+        } else {
+          this.navigate(1)
```
connoratrug, line 212: maybe and note that 1 is the page number

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+      deep: true,
+      immediate: true
+    },
+    value: {
+      handler: function (newVal) {
+        this.val = newVal
+      },
+      deep: true
+    },
+    'value.size' () {
```
connoratrug, line 224: why do we need to watch value.size when value is already watched ( with deep = true) ?

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+    font-weight: 600;
+  }
+}
+</style>
+
+<docs>
+## Basic example
+
+```jsx
+  let model = { size: 20, page: 1 }
```
connoratrug, line 270: How to i know the shape of the model as a user ( maybe add typescript for model) and or describe as part of docs tag

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -0,0 +1,287 @@
...
+    },
+    pageCount () {
+      if (!this.val.count) { return 0 }
+      return Math.floor(parseInt(this.val.count, 10) / this.val.size)
+    },
+    pages () {
+      const pages = []
+
+      const midPage = Math.floor(this.midPageSize / 2)
+      const isMidRange = this.val.page > midPage && !(this.pageCount < this.relativePageSize)
```
connoratrug, line 90: https://sonarcloud.io/project/issues?id=molgenis_molgenis-frontend&pullRequest=464&resolved=false&types=CODE_SMELL

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Renamed variables to descriptive variants

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Added name attribute. I rather use alphabetical order for keys than order of importance.

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Yes, good point. Renamed to `fetchItems`

packages/components-library/src/components/ui/Pagination.vue
jvanveen: It reflects the pagination state. Updated comment to `Reflects the pagination state`

packages/components-library/src/components/ui/Pagination.vue
jvanveen: This can be done in one handler indeed. Checking for changed size now in the value watcher.

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Remove the condition; it wasn't used properly at all

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Added a comment above.

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Added a comment

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Replaced with 'localValue'

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Added as prop

packages/components-library/src/components/ui/Pagination.vue
jvanveen: Added as prop

connoratrug (approved):

jvanveen:
[#7740 Table View: Implement pagination [1pt]](https://trello.com/c/ipMgCo0e/336-7740-table-view-implement-pagination-1pt)

PR #476, opened 2021-01-06
feat(data-explorer): add pagination to de2

jvanveen:
#### Checklist
- [x] Code reviewed


connoratrug:
error on loading 'books' example 
`Uncaught (in promise) TypeError: Cannot read property 'page' of undefined
    at a.<anonymous> (MainView.vue:153)
    at l (runtime.js:45)
    at Generator._invoke (runtime.js:271)
    at Generator.t.<computed> [as next] (runtime.js:97)
    at o (asyncToGenerator.js:5)
    at s (asyncToGenerator.js:27)`

packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -50,12 +50,13 @@ export default {
   methods: {
```
connoratrug, line 50: why does the search component need to be updated when pagination is added ?

packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -50,12 +50,13 @@ export default {
...
     }
   },
   watch: {
     value: {
       handler: function (val, oldVal) {
+        this.searchText = val
+
         if (val === '') {
           this.searchText = ''
         }
```
connoratrug, line 62: this.searchText is set to '' when val === '' , no need to do it again on line 61

packages/data-explorer/src/components/SearchComponent.vue
```diff
@@ -64,12 +65,12 @@ export default {
...
     searchText: {
       // Add handler to support html5 clear search action
-      handler: function (val, oldVal) {
+      handler: function (val) {
         if (val === '') {
           this.handleSearchAction()
         }
       },
-      immediate: true
+      immediate: false
```
connoratrug, line 73: why is searchText no longer evaluated immediate ?

packages/data-explorer/src/repository/dataRepository.ts
```diff
@@ -113,32 +113,41 @@ const getTableDataDeepReference = async (
...
   const response = (await client.get<DataApiResponse>(request)).data
-  const result = { items: response.items.map((item: DataApiResponseItem) => levelNRowMapper(item)) }
-  return result
+
+  const items = response.items.map((item: DataApiResponseItem) => levelNRowMapper(item))
+  return { items, response }
 }
 
-const getTableDataWithLabel = async (tableId: string, metaData: MetaData, columns: string[], rsqlQuery?: string, dataDisplayLimit?: Number) => {
+const getTableDataWithLabel = async (tableId: string, metaData: MetaData, columns: string[], rsqlQuery?: string, pagination?: Object) => {
```
connoratrug, line 127: would suggest to type pagination object

packages/data-explorer/src/store/actions.ts
```diff
@@ -56,7 +55,23 @@ export default {
       commit('applyBookmark')
     }
   },
-  fetchCardViewData: async ({ commit, state, getters }: { commit: any, state: ApplicationState, getters: any }) => {
+  async fetchViewData (store, { tableName, pagination }) {
+    if (store.state.tableName !== tableName) {
+      await store.dispatch('fetchTableMeta', { tableName })
+      if (store.getters.isUserAuthenticated) {
+        await store.dispatch('header/fetchBreadcrumbs', null, { root: true })
```
connoratrug, line 62: in general passing 'null' seems like a code smell, in this case not passing the getters seems to lead to breaking the breadcrumbs

packages/data-explorer/src/store/actions.ts
```diff
@@ -68,34 +83,32 @@ export default {
...
+      request = await dataRepository.getTableDataWithLabel(state.tableName, state.tableMeta, columns, rsqlQuery, pagination)
     }
+
     if (getters.filterRsql === rsqlQuery) {
       // retrieved results are still relevant
-      commit('setTableData', tableData)
+      commit('setTableData', request)
     }
+
+    return request
```
connoratrug, line 109: why is the request itself exposed to the outside, it would make more sense to call the result of fetching the tabel data 'tableData' ( as before) and using the store to manage the data. 

Exposing the low-level request-result to the caller leads to the issue where the caller does not property handle the result resulting in the null pointer that can be seen using the preview

packages/data-explorer/src/store/actions.ts
```diff
@@ -109,16 +122,18 @@ export default {
...
       rsqlQuery,
-      state.dataDisplayLimit)
+      pagination)
     if (getters.filterRsql === rsqlQuery) {
       // retrieved results are still relevant
-      commit('setTableData', tableData)
+      commit('setTableData', request)
     }
+
+    return request
```
connoratrug, line 136: same issue as mentioned above

packages/data-explorer/src/store/state.ts
```diff
@@ -26,8 +25,7 @@ const state: ApplicationState = {
     selections: {}
   },
   searchText: '',
-  bookmark: '',
-  componentRoute: false
+  bookmark: ''
 }
```
connoratrug, line 29: suggest to add page model to store

packages/data-explorer/src/views/MainView.vue
```diff
@@ -33,9 +33,21 @@
...
+        <!--
+          (!) This pagination component is always rendered, because it is
+          responsible for the data fetching in the background. It is only
+          visible in the CardView when the amount of items justifies an
+          extra pagination ui at the bottom of the page.
+        -->
+        <pagination
+          v-show="(dataDisplayLayout === 'TableView') || (!pagination.loading && itemsCount > pagination.size)" class="mt-2"
+          v-model="pagination"
+          :fetchItems="paginateEntities"
```
connoratrug, line 49: why pass a function instead of having a event call the function ( or action)

packages/data-explorer/src/views/MainView.vue
```diff
@@ -133,34 +143,31 @@ export default Vue.extend({
...
-          this.fetchBreadcrumbs()
-        }
-        this.setTableName(tableName)
-      }
-      if (this.dataDisplayLayout === 'CardView') {
-        this.fetchCardViewData()
+    async paginateEntities () {
+      // The data-API uses zero-based page numbering, where pagination is one-based.
+      const pagination = { ...this.pagination, page: this.pagination.page - 1 }
+      const request = await this.fetchViewData({ tableName: this.$route.params.entity, pagination })
```
connoratrug, line 149: could we move this into the store

packages/data-explorer/src/components/SearchComponent.vue
jvanveen: Reverted changes. Doesn't seem to do much.

packages/data-explorer/src/views/MainView.vue
```diff
@@ -133,34 +143,31 @@ export default Vue.extend({
...
+      if (!request || !request.items.length) {
+        this.itemsCount = 0
       } else {
-        this.fetchTableViewData()
+        this.itemsCount = request.response.data.page.totalElements
```
connoratrug, line 153: possible null pointer ( see preview , books example)

packages/data-explorer/src/components/SearchComponent.vue
jvanveen: Reverted changes

packages/data-explorer/src/components/SearchComponent.vue
jvanveen: Reverted changes

packages/data-explorer/src/repository/dataRepository.ts
jvanveen: Yes, see https://github.com/molgenis/molgenis-frontend/pull/477

packages/data-explorer/src/store/actions.ts
jvanveen: This also works:
```
await store.dispatch('header/fetchBreadcrumbs')
```

packages/data-explorer/src/store/actions.ts
jvanveen: Refactored to add page to tableData

packages/data-explorer/src/store/state.ts
jvanveen: Good idea, done.

packages/data-explorer/src/views/MainView.vue
jvanveen: Replaced the fetchItems method with fetchViewData. No need for this method at all.

packages/data-explorer/src/views/MainView.vue
jvanveen: Should be solved with the prior changes

connoratrug (changes requested):
Issue with filter: 

Look at preview ( or local) , signin as admin, open type test entity in data-explorer 2
Use the change filter option to add the 'xbool' filter ( or other bool filter)
Select the 'yes' option in the filter

Expected: rows/cards are filtered and active filter is shown
Actual:  rows/cards are filtered  but no  active filter is shown, if you open the debugger you see a null pointer error 

note: the possible boolean options are of type boolean ( true, false), the activeFilter options that get passed are strings ('true', 'false'), these should have been booleans.

note: issue seems to be bookmarkMapper in data-explorer does not decode string-bool-value to boolean

connoratrug (approved):

PR #503, opened 2021-01-21
fix: Only show description if description is set

connoratrug:
- Add missing check to showing description tooltip in case user in not signed in
- If the (optional) fields descrition is not set, no tooltip should be added

#### Checklist
- [x] Code reviewed
-  User documentation updated
-  No warnings during install
-  Updated javascript typing


connoratrug:
[Fix empty tooltip in case of not signed in and no description](https://trello.com/c/84UJZazo/385-fix-empty-tooltip-in-case-of-not-signed-in-and-no-description)

MaxPostema (approved):

PR #511, opened 2021-01-28
fix(data-explorer): set correct components lib version

connoratrug:
- lerna  bump component lib version on component lib release ( but does not update the lock file)
- this data explorer should use 1.x version as pagination comp has breaking change

#### Checklist
- [ ] Code reviewed


connoratrug:
[fix data-explorer component lib version](https://trello.com/c/CBY2l6aR/396-fix-data-explorer-component-lib-version)

PR #512, opened 2021-01-29
Fix/data explorer route sync issues

connoratrug:
note: ran e2e locally ( passing) due to Jenkins issue 

#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/components-library/src/components/ui/Pagination.vue
+++ b/packages/components-library/src/components/ui/Pagination.vue
@@ -178,18 +178,8 @@ export default {
      * not feasible to detect size property changes from
      * the deep value watcher (See https://vuejs.org/v2/api/#vm-watch)
      */
-    'value.size' (newSize, oldSize) {
-      let currentPosition = this.localValue.page * oldSize
-      // Keep in mind that with 201 items, the current page may be 3 when
-      // showing 100 items per page. When switching back to 50 items per
-      // page, there are only 5 pages left. Make sure the currentPosition
-      // never exceeds the actual count.
-      if (currentPosition > this.localValue.count) {
-        currentPosition = this.localValue.count
-      }
-
-      const currentNewPage = Math.ceil(currentPosition / newSize)
-      this.updateValue({ page: currentNewPage })
+    'value.size' () {
+      this.updateValue({ page: 1 })
     }
   }
 }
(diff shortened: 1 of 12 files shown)


connoratrug:
[#7228 As explorer user I can choose what column to sort so I can more easily go through the data [2]](https://trello.com/c/Z0vveJBB/365-7228-as-explorer-user-i-can-choose-what-column-to-sort-so-i-can-more-easily-go-through-the-data-2)

packages/components-library/tests/unit/filters/containers/FilterContainer.spec.ts
```diff
@@ -79,4 +79,9 @@ describe('FilterContainer.vue', () => {
     await wrapper.find('.btn.btn-block.btn-primary').trigger('click')
     expect(wrapper.find('.modal-dialog')).toBeDefined()
   })
+
+  it('reacts to filtersShown changes', async () => {
+    await wrapper.setProps({ filtersShown: ['color'] })
```
jelmerveen, line 84: better use includes, to make the test more robust?

packages/components-library/tests/unit/filters/containers/FilterContainer.spec.ts
```diff
@@ -79,4 +79,9 @@ describe('FilterContainer.vue', () => {
...
+    expect((wrapper.vm).listOfVisibleFilters[0].label).toBe('Color')
```
jelmerveen, line 85: maybe map the array of objects to an array of labels and then use includes('Color') so if, for some instance the order changes the test won't break

jelmerveen (approved):
Approved, but maybe check my comment.

packages/components-library/src/components/ui/Pagination.vue
```diff
@@ -178,18 +178,8 @@ export default {
...
-      // page, there are only 5 pages left. Make sure the currentPosition
-      // never exceeds the actual count.
-      if (currentPosition > this.localValue.count) {
-        currentPosition = this.localValue.count
-      }
-
-      const currentNewPage = Math.ceil(currentPosition / newSize)
-      this.updateValue({ page: currentNewPage })
+    'value.size' () {
+      this.updateValue({ page: 1 })
```
jvanveen, line 182: This only triggers a data update when a user is changing the page size from a page that is not 1.

jvanveen (changes requested):

PR #520, opened 2021-02-04
feat(data-explorer): Allow user to hide columns

connoratrug:
- Add 'hide' query-param to route type
- Set hiddenColumns on state when set in route
- Hide hidden columns in tableView
- Hide hidden fields in defaultCardView
- Do NOT hide anything in custom cardview

#### Todo
- [x] Add hide to route
- [x] Update store with route
- [x] Reflect in view
- [x] Add UI for hiding columns ( toolbar-btn to modal , maybe combine with col-header-icon-btn)   

#### Checklist
- [x] Code reviewed
-  User documentation updated
- No warnings during install
- Updated javascript typing

### DIFF
--- a/packages/data-explorer/src/components/ColumnSelection.vue
+++ b/packages/data-explorer/src/components/ColumnSelection.vue
@@ -17,17 +17,8 @@
 export default {
   computed: {
     allColumns () {
-      if (!this.$store.state.tableData) {
-        return []
-      }
-
-      return [
-        ...new Set( // creating a set, to deduplicate
-          this.$store.state.tableData.items
-            .map((item) => Object.keys(item)) // get all properties from tableData
-            .reduce((prev, current) => prev.concat(current)) // do simple flatmap
-        )
-      ]
+      const { tableMeta } = this.$store.state
+      return tableMeta ? tableMeta.attributes.map((item) => item.name) : []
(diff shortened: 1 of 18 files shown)


connoratrug:
[#2887 As explorer user I want to quickly show/hide attributes (columns in Table, fields in Cards) [3]](https://trello.com/c/Jyruu8sv/361-2887-as-explorer-user-i-want-to-quickly-show-hide-attributes-columns-in-table-fields-in-cards-3)

packages/data-explorer/src/components/ColumnSelection.vue
```diff
@@ -0,0 +1,56 @@
...
+      <span class="ml-2" v-text="column"></span>
+    </label>
+  </div>
+</template>
+
+<script>
+export default {
+  computed: {
+    allColumns () {
+      return this.$store.state.tableData
```
connoratrug, line 20: general architecture is to not use the store in the components, but props down and events up, then use the components in the Views and use the View to wire it to the store

jelmerveen: then we need to have more code in the component, to watch for changes. But can be done

MaxPostema (approved):

PR #535, opened 2021-02-16
feat: #7888 add filter tab to toggle FiltersView

jvanveen:
#### Checklist
- [ ] Functionality works & meets specifications
- [ ] Code reviewed
- [ ] Code unit/integration/system tested
- [ ] User documentation updated
- [ ] Conventional commits (squash if needed)
- [ ] No warnings during install
- [ ] Updated javascript typing

PR #537, opened 2021-02-16
feat: control filter sidebar visibility from querystring

jvanveen:
#### Checklist
- [x] Code reviewed
-  User documentation updated
-  No warnings during install
-  Updated javascript typing


jvanveen:
De knop in de toolbarview verdwijnt met https://github.com/molgenis/molgenis-frontend/pull/535, dus die is ongewijzigd in deze pr.

connoratrug:
[add filter show/hide status to the route query (jeroen)  https://github.com/molgenis/molgenis-frontend/pull/537](https://trello.com/c/kPWwLJbM/418-add-filter-show-hide-status-to-the-route-query-jeroen-https-githubcom-molgenis-molgenis-frontend-pull-537)

connoratrug (commented):
When i close the filter bar , when open the bar , when click close again nothing happens.
When i click close a second time the bar does close

packages/data-explorer/src/router.ts
```diff
@@ -1,34 +1,26 @@
...
-      component: MainView,
-      name: 'main-view'
-    },
-    { path: '/:entity?',
-      redirect: to => {
-        let entity = 'sys_ts_DataExplorerEntitySettings'
-        if (to.params.entity) {
-          entity = to.params.entity
-        }
+const routes = [
```
connoratrug, line 3: should the file name be changed to routes ?

packages/data-explorer/src/types/RouteQuery.ts
```diff
@@ -3,5 +3,6 @@ export type RouteQuery = {
     size?: string, // page size integer as a string
     sort?: string, // name of sort column, prefix with '-' for reverse sort
     filter?: string // hashed filter object
+    filterbar?: '0' | '1' // filter sidebar
```
connoratrug, line 6: why is an int (as string) instead of a bool ( as string) , why filterbar and not 'showfilters' or 'isfilersshown' or something like that

packages/data-explorer/src/views/FiltersView.vue
```diff
@@ -90,6 +89,13 @@ export default Vue.extend({
         name: 'main-view',
         query: { ...this.$route.query, filter: this.compressedRouteFilter }
       })
+    },
+    toggleFilterbar () {
+      let filterbarState = this.$route.query.filterbar ? parseInt(this.$route.query.filterbar, 10) : 1
```
connoratrug, line 94: why the parsing instead of a simple bool

connoratrug (dismissed):
the need to click two times needs to be fixed i thing , the rest is working nicely

packages/data-explorer/src/router.ts
jvanveen: That makes sense indeed. Changed

packages/data-explorer/src/types/RouteQuery.ts
jvanveen: I suppose its a matter a of personal preference. I find an int easier to parse to a boolean and to toggle states between. Same for the naming of the variable; I would rather not use verbs, because the other parameters are nouns.

packages/data-explorer/src/views/FiltersView.vue
jvanveen: The querystring cannot be typed differently than string (afaik). Flipping an int seems more straightforward to me than switching between 'true' and 'false' strings.

jvanveen:
Renamed the filterbar variable to hideSidebar (same as the state variable), using 'false'/'true' now instead of '0'/'1'. Renamed router.ts to 'routes.ts'. Fixed tests.

MaxPostema (approved):

MaxPostema (approved):

PR #542, opened 2021-02-19
chore(data-explorer): initial combination of equal logic

jelmerveen:
#### Checklist
- [ ] Code reviewed

### DIFF
--- a/packages/data-explorer/src/store/actions.ts
+++ b/packages/data-explorer/src/store/actions.ts
@@ -4,7 +4,7 @@ import * as metaDataRepository from '@/repository/metaDataRepository'
 import * as dataRepository from '@/repository/dataRepository'
 import * as metaDataService from '@/repository/metaDataService'
 import * as metaFilterMapper from '@/mappers/metaFilterMapper'
-import router from '@/router'
+import { Store } from 'vuex'
 
 export default {
   fetchTableMeta: async ({ commit, getters, dispatch, state }, payload: { tableName: string }) => {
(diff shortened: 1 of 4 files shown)


packages/data-explorer/src/store/actions.ts
```diff
@@ -26,45 +26,52 @@ export default {
...
+    if (state.dataDisplayLayout === 'CardView') {
+      dispatch('fetchCardViewData')
+    } else {
+      dispatch('fetchTableViewData')
+    }
+  },
+  /**
+   * Don't call this function directly (except for tests), use fetchViewData instead
+   */
+  fetchCardViewData: async ({ commit, state, getters }) => {
```
connoratrug, line 50: the do not export it

packages/data-explorer/src/store/actions.ts
```diff
@@ -4,7 +4,7 @@ import * as metaDataRepository from '@/repository/metaDataRepository'
 import * as dataRepository from '@/repository/dataRepository'
 import * as metaDataService from '@/repository/metaDataService'
 import * as metaFilterMapper from '@/mappers/metaFilterMapper'
-import router from '@/router'
+import { Store } from 'vuex'
```
connoratrug, line 7: store is never used ..

packages/data-explorer/src/store/actions.ts
```diff
@@ -26,45 +26,52 @@ export default {
...
+    }
+  },
+  /**
+   * Don't call this function directly (except for tests), use fetchViewData instead
+   */
+  fetchCardViewData: async ({ commit, state, getters }) => {
+    const { tableName, tableMeta, tablePagination, sort, tableSettings, dataDisplayLayout } = state
+
+    // failsafe measures.
+    if (!tableName) return
```
connoratrug, line 54: this is an error state , why would you return null

packages/data-explorer/src/store/actions.ts
```diff
@@ -26,45 +26,52 @@ export default {
...
+      dispatch('fetchCardViewData')
+    } else {
+      dispatch('fetchTableViewData')
+    }
+  },
+  /**
+   * Don't call this function directly (except for tests), use fetchViewData instead
+   */
+  fetchCardViewData: async ({ commit, state, getters }) => {
+    const { tableName, tableMeta, tablePagination, sort, tableSettings, dataDisplayLayout } = state
```
connoratrug, line 51: destructuring leads to vars being declared to soon, out side of context and possibly unused

connoratrug (changes requested):
What is the PR trying to achieve ? What problem is it solving, and what problems is it introducing

packages/data-explorer/src/store/actions.ts
jelmerveen: Destructuring gives us more clarity upfront and removes unnecessary trailing paths, it's aready happening in the function.
unnecessary vars isn't an issue, as you only export what you need. And if there would be a rogue variable, the IDE will tell you so.

jelmerveen:
This PR was about streamlining the dataflow through fetchViewData, as all the views are using this.
BUT there were remnants to the other ones inside vue files that made it look like the functions are all called seperately. Which was not the case and caused confusion when fixing a bug. We decided to make a seperate pull request to tackle this issue, by using the fetchViewData at all times and rely on the vuex state on the correct path. Which makes it cleaner, more readable and less prone to bugs.

Examples are: packages/data-explorer/src/views/ClipboardView.vue  
  ...mapActions(['fetchTableViewData']), is imported but never used. Suggesting it should be called directly.

And again in packages/data-explorer/src/views/TableView.vue

PR #547, opened 2021-02-24
feat: #7889 merged data-row-edit in de2

jvanveen:
#### Checklist
- [x] Code reviewed


jvanveen:
[#7889 (TBD: As developers we want to combine DE and Data Row Editor so that we can simplify the build/dependencies)](https://trello.com/c/rIowzUKW/410-7889-tbd-as-developers-we-want-to-combine-de-and-data-row-editor-so-that-we-can-simplify-the-build-dependencies)

MaxPostema:
I don't think the "add" in DE2 works correctly there is still a page reload. Editing a row feels nice and fast now! With routing like this we can still use it outside of DE2 (from a app or something) it does feel a bit slower when using it from outside of DE2, but that could be our finicky master server.

I think this works wel. I wonder what the other team members think.

jvanveen:
- Fixed the add button 
- Changed the navigator default link to point to the new router link
- Applied route guards to make sure that all query parameters are filled when entering the MainView
- Applied loading state, so the old data doesn't show up twice after loading another entity because of changing meta-data
- Fixed tests
- Fixed test warnings (use ts 3.9.3 with latest ts-jest)
- Fixes eslint warnings

packages/data-explorer/src/main.ts
```diff
@@ -74,18 +74,32 @@ Vue.config.productionTip = false
...
+    } else { next() }
+  } else { next() }
+})
+
 Vue.use(i18n, {
-  lng: 'en',
-  fallbackLng: 'en',
-  namespace: ['dataexplorer'],
+  lng,
+  fallbackLng,
```
MaxPostema, line 100: What happens if window.__INITIAL_STATE__ is empty. Maybe we need a in application fallback 'en'?

packages/data-explorer/src/main.ts
```diff
@@ -74,18 +74,32 @@ Vue.config.productionTip = false
 // This way the mainView can coordinate the events
 Vue.prototype.$eventBus = new Vue()
 
+// @ts-ignore
+const { lng, fallbackLng } = window.__INITIAL_STATE__
```
MaxPostema, line 78: Maybe add this line before line #98 "Vue.use" to group it together?

packages/data-explorer/public/index.html
```diff
@@ -7,6 +7,13 @@
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>data-explorer</title>
     <script>
+    window.__INITIAL_STATE__ = {
```
MaxPostema, line 10: Having a static __INITIAL_STATE__ feels broken. Can we get thees values from context api?

packages/data-explorer/src/components/DataRowEdit.vue
```diff
@@ -0,0 +1,267 @@
...
+    },
+    parent: {
+      type: Object,
+      required: false,
+      default: null
+    }
+  },
+  data () {
+    return {
+      dataExplorerBaseUrl: window.__INITIAL_STATE__.dataExplorerBaseUrl,
```
MaxPostema, line 116: is this in the context api?

packages/data-explorer/src/routes.ts
```diff
@@ -1,26 +1,36 @@
...
+export const routes = [
   {
-    path: '/:entity/:view',
+    path: '/explorer/:entity',
     component: MainView,
-    name: 'main-view'
+    name: 'de-view'
+  },
+  {
+    path: '/explorer/:entity/create',
```
MaxPostema, line 13: Did you not say that you wont change the urls in this commit?

packages/data-explorer/src/store/actions.ts
```diff
@@ -26,15 +26,14 @@ export default {
...
-  async fetchViewData ({ dispatch, state }: any) {
+  async fetchViewData ({ commit, dispatch, state }: any) {
     if (state.dataDisplayLayout === 'CardView') {
       dispatch('fetchCardViewData')
     } else {
       dispatch('fetchTableViewData')
     }
   },
   fetchCardViewData: async ({ commit, state, getters }: { commit: any, state: ApplicationState, getters: any }) => {
-    commit('setLoading', true)
```
MaxPostema, line 37: Why remove the spinner?

packages/data-explorer/src/views/MainView.vue
```diff
@@ -139,21 +139,23 @@ export default Vue.extend({
...
     this.setRouteQuery(this.$route.query)
-    this.setDataDisplayLayout(this.$route.params.view)
+    this.setDataDisplayLayout(this.$route.query.view)
     this.fetchViewData()
   },
   destroyed () {
     this.$eventBus.$off('delete-item')
   },
   async beforeRouteUpdate (to, from, next) {
+    this.setLoading(true)
```
MaxPostema, line 149: Is this beter here that in the actions?

MaxPostema:
Schermafbeelding 2021-03-05 om 13 44 51

I think the title is to close to the menu bar. (chrome)

packages/data-explorer/src/main.ts
jvanveen: Good one! Restored the original situation where lng and fallbackLng were set in main.ts

packages/data-explorer/src/main.ts
jvanveen: I removed them together with the __INITIAL__STATE in index.html

packages/data-explorer/src/components/DataRowEdit.vue
jvanveen: The 'dataExplorerBaseUrl' is actually unused in DataRowEdit. I removed the whole data property.

packages/data-explorer/src/routes.ts
jvanveen: Yes, that was to keep the link in the navigator working. I fixed the link in the navigator now instead.

packages/data-explorer/src/views/MainView.vue
jvanveen: I think so! :) By putting the loading indicator here, its easier to spot where the loading occurs and where it ends.

packages/data-explorer/src/store/actions.ts
jvanveen: Its wrapped around the whole data fetch function, including the fetchTablePermissions & fetchTableMeta

packages/data-explorer/src/components/DataRowEdit.vue
```diff
@@ -0,0 +1,266 @@
...
+        }
+      } else {
+        this.saveFailed = true
+      }
+    },
+    onCancelClick () {
+      this.parent ? this.showParent() : this.goBackToPluginCaller()
+    },
+    goBackToPluginCaller () {
+      window.history.go(-1)
```
connoratrug, line 191: should probably vue the router now it it part of the data explorer , ```window.history
``` was more of spike workaround

File: packages/data-explorer/src/components/DataRowEdit.vue [hunk 9]
```diff
@@ -0,0 +1,266 @@
...
+        showNonVisibleAttributes: true,
+        mapperMode: dataRowId ? 'UPDATE' : 'CREATE',
+        booleanLabels: {
+          trueLabel: this.$t('data-row-edit-boolean-true'),
+          falseLabel: this.$t('data-row-edit-boolean-false'),
+          nillLabel: this.$t('data-row-edit-boolean-null')
+        }
+      }
+      try {
+        const resp = await repository.fetch(entity, dataRowId)
```
connoratrug, line 245: should probably be moved to store to be consistent with data explorer architecture

packages/data-explorer/src/components/DataRowEdit.vue
```diff
@@ -0,0 +1,266 @@
...
+      this.parent.setRef(false) // show parent
+      this.parent.$refs.refContainer.removeChild(this.$el) // destroy child
+    },
+    handleError (error) {
+      const alertMsg = this.errorToMessage(error)
+      this.showForm = true
+      this.isSaving = false
+      this.alert = { message: alertMsg, type: 'danger' }
+    },
+    errorToMessage (error) {
```
connoratrug, line 209: should probably be replaced by more generic data explorer error handler

packages/data-explorer/src/components/DataRowEdit.vue
jvanveen: Refactoring the DataRowEdit component is out of scope for this ticket.

packages/data-explorer/src/components/DataRowEdit.vue
jvanveen: Refactoring the DataRowEdit component is out of scope for this ticket.

packages/data-explorer/src/components/DataRowEdit.vue
jvanveen: Refactoring the DataRowEdit component is out of scope for this ticket.

packages/data-explorer/src/main.ts
```diff
@@ -74,18 +74,29 @@ Vue.config.productionTip = false
...
+// Navigation guard that makes sure that all explorer
+// query params are present.
+router.beforeEach((to, from, next) => {
+  globalThis.defaultRouteQuery = defaultRouteQuery
+  globalThis.to = to.query
+  if (to.name === 'de-view') {
+    const missingQueryParams = Object.keys(defaultRouteQuery).filter((i) => !Object.keys(to.query).includes(i))
+    if (missingQueryParams.length) {
+      next({ path: `/explorer/${to.params.entity}`, query: defaultRouteQuery })
+    } else { next() }
```
connoratrug, line 91: why is this code needed , it looks like it trying to hide a issue that should be fixed at the source

jvanveen: It verifies whether all query params are available when it enters the de-view route. Otherwise it redirects with all the required query params.

packages/data-explorer/src/routes.ts
```diff
@@ -1,26 +1,36 @@
...
+    props: true,
+    name: 'de-create',
+    component: DataRowEdit
   },
   {
-    path: '/:entity?',
+    path: '/explorer/:entity/:dataRowId/edit',
+    props: true,
+    name: 'de-edit',
```
connoratrug, line 21: why the 'de' prefix , suggest to replace with 'edit'

packages/data-explorer/src/routes.ts
```diff
@@ -1,26 +1,36 @@
...
-const routes = [
+export const defaultRouteQuery:any = { page: 1, size: 20, hideSidebar: 'false', view: 'CardView' }
+
+export const routes = [
   {
-    path: '/:entity/:view',
+    path: '/explorer/:entity',
     component: MainView,
-    name: 'main-view'
+    name: 'de-view'
```
connoratrug, line 10: why the 'de'; prefix , the whole app is in de data-explorer namespace , main-view matches the view name and seems more descriptive

packages/data-explorer/src/routes.ts
jvanveen: Its useful to prefix actions with a module, so that you don't get name collisions when merging other modules (e.g. the navigator or data-row-edit)

packages/data-explorer/src/routes.ts
jvanveen: See comment above.

packages/data-explorer/src/types/ApiResponse.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-use-before-define */
```
connoratrug, line 1: this seems like a issue that should no be suppressed

packages/data-explorer/src/types/RouteQuery.ts
```diff
@@ -5,4 +5,5 @@ export type RouteQuery = {
     filter?: string // hashed filter object
     hideSidebar?: 'false' | 'true'
     hide?: string // comma separated list of column names
+    view?: string
```
connoratrug, line 8: could replace with 'card' | 'table'

packages/data-explorer/src/types/ApiResponse.ts
jvanveen: The downgrade to typescript 3.9.3 seemed to fix the eslint parser, but I didn't feel like fixing all the violations in this file. They all rely on each other and its easy to overlook a definition.

packages/data-explorer/src/types/RouteQuery.ts
jvanveen: Good suggestion, updated!

packages/data-explorer/src/components/DataRowEdit.vue
connoratrug: i do not agree , as it now becomes past of the whole DE2 data flow

packages/data-explorer/src/routes.ts
connoratrug: its not a module but a view ( or should be a view)

connoratrug (commented):
The PR in its current state merges the data-row-edit code into the data-explorer package, by not integrating it into the data-explorer app architecture a lot of technical dept is added. Would suggest to move 'data-fetch', 'routing' add 'error handling' part of data-row-edit into the data-explorer to avoid different code/architecture styles in a single app. Once merged these are harder to find and update.

packages/data-explorer/src/main.ts
connoratrug: so then it hides the error making it harder to fix, if a single param is missing , the whole thing now gets replaced with the default , this seems incorrect

MaxPostema (approved):

PR #551, opened 2021-02-26
Feat/de2 file type

connoratrug:
#### Checklist
- [ ] Code reviewed
- User documentation updated


connoratrug:
[File type de2](https://trello.com/c/ndC7gUiR/413-file-type-de2)

MaxPostema (approved):
I Like it. I think the cherry on top would be if you would give the size property a unit.. I think the number is in byte's? Now it's a number without meaning something.

PR #558, opened 2021-03-07
Feat/template editor

connoratrug:
Basic template editor

Open any entity in dataexplorer and add ```/editor``` to the path to try it out

- [x] Live preview ( with item selector)
- [x] Save template
- [x] Show template props tree
- [ ] Save template props
- [ ] Errors in view ( instead of console)
- [ ] Default preview template, when no template is set
- [ ] Fix deep reference stuff ( can it be fixed ?)
- [ ] Extract reusable components ( its one bug vue file)
- [ ] Test code

#### Checklist
- [ ] Code reviewed


connoratrug:
spike , no need to keep PR open , leave as idea

PR #561, opened 2021-03-09
feat: checkboxes are now able to be filtered

jelmerveen:
now you can add an array of values to the checkbox filter and it will only show those options

#### Checklist
- [ ] Code reviewed


jvanveen (commented):
Can't this be solved within the options method?
It seems to add complexity by introducing yet another way of filtering, next to sliceOptions/resolvedOptions?

PR #656, opened 2021-06-24
fix(components-library): abstracted satisfy-all and changed it's styling

jelmerveen:
#### Checklist
- [x] Code reviewed
- [x] Add to release notes

### DIFF
--- a/packages/components-library/src/components/filters/CheckboxFilter.vue
+++ b/packages/components-library/src/components/filters/CheckboxFilter.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
-    <div v-if="showSatisfyAllCheckbox" class="query-type-selector">
-      <label class="label-disabled">
+    <div v-if="showSatisfyAllCheckbox" class="text-right mb-2">
+      <label>
         {{ satisfyAllLabel }}
         <b-form-checkbox
           v-model="satisfyAll"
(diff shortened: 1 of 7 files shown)


packages/components-library/src/components/filters/CheckboxFilter.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
-    <div v-if="showSatisfyAllCheckbox" class="query-type-selector">
-      <label class="label-disabled">
+    <div v-if="showSatisfyAllCheckbox" class="text-right mb-2">
+      <label>
```
fdlk, line 4: why remove disabled class here and not in the other checkbox?

packages/components-library/src/components/filters/CheckboxFilter.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
-    <div v-if="showSatisfyAllCheckbox" class="query-type-selector">
-      <label class="label-disabled">
+    <div v-if="showSatisfyAllCheckbox" class="text-right mb-2">
```
fdlk, line 3: A special class allows users of the component to override the styling, which is sort of nice I think. How about keeping some kind of class label?

fdlk (changes requested):

packages/components-library/src/components/filters/CheckboxFilter.vue
jelmerveen: would like to defer this change untill needed

fdlk (approved):

PR #696, opened 2021-08-30
feat: data-row-permissions app

jelmerveen:
#### Checklist
- [x] Code reviewed
- [ ] Add to release notes


packages/data-row-permissions/src/components/ChangeOwnership.vue
```diff
@@ -0,0 +1,65 @@
...
+    value (newValue) {
+      this.changeOwner = newValue
+      this.newOwner = this.permissionObject.ownedByUser || this.permissionObject.ownedByRole
+    }
+  },
+  methods: {
+    ...mapActions(['updatePermissions']),
+    save () {
+      const updateOwnership = {}
+      if (this.userOptions.filter(f => f.value === this.newOwner).length) {
```
fdlk, line 51: `this.userOptions.some(f => f.value === this.newOwner)`

packages/data-row-permissions/src/components/DataStatus.vue
```diff
@@ -0,0 +1,58 @@
...
+        <div class="mt-5">
+          {{ $t('data-row-permissions-spinner-text') }}
+        </div>
+      </div>
+      <span
+        v-else-if="isAuthenticated"
+        class="mt-2">{{ $t('data-row-permissions-no-results') }}</span>
+      <span
+        v-else
+        class="mt-2">{{ $t('data-row-permissions-no-results-cta') }}&nbsp;<a href="/login">{{ $t('data-row-permissions-cta-login') }}</a>.</span>
```
fdlk, line 19: Since this is an admin screen, why not open the login dialog right away if user isn't authenticated?

packages/data-row-permissions/src/main.js
```diff
@@ -0,0 +1,67 @@
+import Vue from 'vue'
+import App from './App.vue'
+import router from './router'
+import store from './store'
+
+import i18n from '@molgenis/molgenis-i18n-js'
+import { BootstrapVue } from 'bootstrap-vue'
+
+// Import Bootstrap an BootstrapVue CSS files (order is important)
+import 'bootstrap/dist/css/bootstrap.css'
```
fdlk, line 10: Use molgenis theme instead.

packages/data-row-permissions/src/logic/filter.js
```diff
@@ -0,0 +1,42 @@
...
+      match = match.concat(objectArray.filter(object => object[propArray[i]] &&
+        (object[propArray[i]].toLowerCase().includes(matchOn) ||
+        (searchProps.includes(propArray[i]) && propArray[i].toLowerCase().includes(matchOn)))))
+    }
+
+    if (matches.length) {
+      for (const m of match) {
+        // check if already found
+        if (matches.every(matched => matched[propArray[0]] !== m[propArray[0]])) {
+          matches.push(m)
```
fdlk, line 34: This is a bit awkward combination of filter and imperative code.
Create a filter function that works to check if a single element matches one of the search criteria.
Then filter the object array using that filter function.

packages/data-row-permissions/src/store/actions.js
```diff
@@ -0,0 +1,80 @@
...
+    api.get(`/api/permissions/${state.route.params.entityId}/${state.route.params.objectId}`).then((response) => {
+      commit('setPermissionObject', response)
+      commit('setResponseStatus', response)
+      commit('setStartInAddMode', response)
+    }).catch(e => {
+      commit('setResponseStatus', e)
+    })
+  },
+  updatePermissions ({ state, dispatch }, changedPermissionObjects) {
+    // TODO upgrade api-client
```
fdlk, line 50: Create an issue in the api client repository and link to it

packages/data-row-permissions/src/components/ServerStatus.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <div
+    v-if="unauthorized">
+    <span>{{ $t('data-row-permissions-no-results-cta') }}&nbsp;<a href="/login">{{ $t('data-row-permissions-cta-login') }}</a>.</span>
```
fdlk, line 4: If the user is unauthorized, explain that this is an admin only feature. Though I think if that happens it is due to a server misconfiguration if the plugin appears in the user's menu and they can only reach it through an external link.

packages/data-row-permissions/src/store/actions.js
```diff
@@ -0,0 +1,80 @@
+import api from '@molgenis/molgenis-api-client'
+
+export const actions = {
+  getAllRoles ({ commit }) {
+    api.get('/api/data/sys_sec_Role').then((response) => {
```
fdlk, line 5: Handle errors for all api calls, i.e. What if the server goes down or the session has timed out?

packages/data-row-permissions/src/views/DataRowPermissions.vue
```diff
@@ -0,0 +1,276 @@
...
+              </td>
+            </tr>
+            <tr
+              v-for="(permission, index) of permissions"
+              :key="permission.id">
+              <td class="align-middle">
+                {{ permission.user || permission.role }}
+              </td>
+              <td class="align-middle w-25">
+                {{ permission.role ? 'Role' : 'User' }}
```
fdlk, line 109: I18n this

packages/data-row-permissions/src/views/DataRowPermissions.vue
```diff
@@ -0,0 +1,276 @@
...
+    return {
+      search: '',
+      addMode: false,
+      editMode: false,
+      deleteMode: false,
+      changeOwner: false,
+      newPermissionType: '',
+      newPermissionObject: {},
+      changedPermissionObjects: [],
+      available_types: [{ text: 'Select a type', value: '' }, { text: 'Role', value: 'role' }, { text: 'User', value: 'user' }]
```
fdlk, line 185: I18n these

packages/data-row-permissions/tests/e2e/specs/test-with-pageobjects.js
```diff
@@ -0,0 +1,30 @@
...
+  beforeEach: (browser) => browser.init(),
+
+  'e2e tests using page objects': (browser) => {
+    const homepage = browser.page.homepage()
+    homepage.waitForElementVisible('@appContainer')
+
+    const app = homepage.section.app
+    app.assert.elementCount('@logo', 1)
+    app.expect.section('@welcome').to.be.visible
+    app.expect.section('@headline').text.to.match(/^Welcome to Your Vue\.js (.*)App$/)
```
fdlk, line 19: I doubt that's still true :)

packages/data-row-permissions/src/views/RlsObjectSelector.vue
```diff
@@ -0,0 +1,87 @@
...
+  computed: {
+    ...mapState(['rlsEntities', 'responseStatus', 'rlsObjects']),
+    currentLabel () {
+      if (this.rlsEntities && this.rlsEntities.length) {
+        return this.rlsEntities.find(entity => entity.id === this.entityId).label
+      }
+
+      return ''
+    },
+    loaded () {
```
fdlk, line 67: consider making this a getter in the store instead

packages/data-row-permissions/vue.config.js
```diff
@@ -0,0 +1,75 @@
+const PROXY_TARGET = 'https://preview-pr-9315-12.dev.molgenis.org/' // 'https://master.dev.molgenis.org'
```
fdlk, line 1: bit chicken and egg what the right order is, but I'd already check in master.dev here because otherwise we need another PR after merging

fdlk (changes requested):
Really nice work!!
Left some nitty-gritty comments but am very enthusiastic!

packages/data-row-permissions/src/logic/filter.js
```diff
@@ -0,0 +1,42 @@
...
+  const numberOfProperties = propArray.length
+  for (let i = 0; i < numberOfProperties; i++) {
+    let match = []
+
+    if (!searchProps.length) {
+      match = objectArray.filter(object => object[propArray[i]] && object[propArray[i]].toLowerCase().includes(matchOn))
+    } else {
+      match = match.concat(objectArray.filter(object => object[propArray[i]] &&
+        (object[propArray[i]].toLowerCase().includes(matchOn) ||
+        (searchProps.includes(propArray[i]) && propArray[i].toLowerCase().includes(matchOn)))))
```
fdlk, line 27: toLowerCase doesn't work nicely with accents. use a case-insensitive regex instead

packages/data-row-permissions/src/logic/filter.js
fdlk: So if I understand the method correctly:
```
export const filterObjectOnStringProperties = (objectArray, propArray, search, searchProps = []) => {
  if (!search) {
    return objectArray
  }
  const matchOn = new RegExp(search, "i")
  return objectArray.filter(object => {
    const stringMatch = propArray
      .map(prop => object[prop])
      .some(value => matchOn.test(value))
    const propMatch = searchProps
      .filter(prop => matchOn.test(prop))
      .some(prop => object[prop])
    return stringMatch || propMatch
  })
}
```

packages/data-row-permissions/tests/e2e/specs/test-with-pageobjects.js
jelmerveen: it's just boilerplate for later.

packages/data-row-permissions/src/logic/filter.js
jelmerveen: this seems to work nicely? "OØÁÄÆÅÉ".toLowerCase()

packages/data-row-permissions/src/components/DataStatus.vue
jelmerveen: technically it's not admin only, it's for owners of objects. So little extra hints aren't that bad?

packages/data-row-permissions/src/main.js
jelmerveen: Molgenis theme does work (have tested) It will overwrite

packages/data-row-permissions/vue.config.js
jelmerveen: good catch, although not breaking :) I'll change it back!

packages/data-row-permissions/src/views/RlsObjectSelector.vue
jelmerveen: done

fdlk (approved):

